PR #5, opened 2026-04-05
Potential fix for code scanning alert no. 18: Uncontrolled command line

cdeust:
Potential fix for [https://github.com/cdeust/Cortex/security/code-scanning/18](https://github.com/cdeust/Cortex/security/code-scanning/18)

General approach: Validate and normalize the user-supplied filename before it is used to build git command arguments. Enforce that the resolved path is inside the git repository, disallow path traversal components (`..`), NUL bytes, and empty paths, and only allow git to see repo-relative paths. If validation fails, treat it as “file not found” rather than passing the malicious value to subprocess.

Best concrete fix with minimal behavior change:

1. Strengthen `_to_relative` in `mcp_server/infrastructure/git_diff.py` to:
   - Strip NUL bytes.
   - Reject empty names.
   - For absolute paths, only return a relative path if they cleanly resolve under `git_root`; otherwise return `""` (invalid).
   - For non-absolute paths, normalize (`os.path.normpath`) relative to `git_root` and ensure the result stays inside `git_root` (no path traversal); otherwise return `""`.
   - In all error cases, return `""` instead of the raw input.

2. In `resolve_file`, handle the case where `_to_relative` returns `""` by immediately returning `None` so that `rel` is never an unsafe value passed to `_git_cmd`.

3. In `_git_cmd`, add a small guard that if the `cmd` list contains a path argument that is an empty string or contains a NUL character, it bails out early and returns `""`. This is mostly belt-and-suspenders; the main control is in `_to_relative`.

These changes keep existing functionality for valid, in-repo paths (absolute or relative) while blocking malicious or nonsensical paths from reaching `subprocess.run`. No changes are required to `http_standalone.py` or `http_file_diff.py`, because the validation happens in the infrastructure layer where the untrusted path first interacts with git and the filesystem.

Concretely, all edits are in `mcp_server/infrastructure/git_diff.py`:
- Replace the body of `_to_relative` with a version that strictly resolves paths under `git_root` and returns `""` for invalid inputs.
- Add a guard at the beginning of `resolve_file` to handle `rel == ""`.
- Add a small validation check at the top of `_git_cmd` to defensively reject obviously invalid arguments (such as ones containing `\x00`).

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/mcp_server/infrastructure/git_diff.py
+++ b/mcp_server/infrastructure/git_diff.py
@@ -43,6 +43,9 @@ def resolve_file(name: str, git_root: Path) -> str | None:
     Uses git ls-files for verification.
     """
     rel = _to_relative(name, git_root)
+    if not rel:
+        # Invalid or unsafe path (outside repo, traversal, etc.)
+        return None
 
     # Check if it's tracked by git
     tracked = _git_cmd(["git", "ls-files", "--", rel], git_root)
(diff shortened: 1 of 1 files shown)


cdeust:
Fixed in main: resolved all CodeQL alerts including command injection in git_diff.py in commit 2326394

Issue #18, opened 2026-05-05
query_methodology: cwd doesn't slug-normalize Windows absolute paths to existing domain


PSGSupport:
## Summary

`query_methodology(cwd=\"C:/Users/michael.crawford\")` returns a hollow profile (`coldStart=false`, but every profile field empty) even when `list_domains` shows the matching domain exists with 18 sessions and confidence 0.36.

## Reproducer

```
list_domains() →
  [{id: \"c--users-michael-crawford\", sessions: 18, confidence: 0.36, ...}, ...]

query_methodology(cwd: \"C:/Users/michael.crawford\") →
  { domain: null,
    coldStart: false,
    style: {},
    entryPoints: [],
    recurringPatterns: [],
    blindSpots: [],
    message: \"Domain \\\"None\\\" detected but no profile built yet\" }

query_methodology(project: \"c--users-michael-crawford\") →
  { domain: \"c--users-michael-crawford\",
    hotMemories: [...],   # populated
    style: {},            # still empty
    entryPoints: [],      # still empty
    ... }
```

So there are actually two related problems:

1. The cwd→domain classifier doesn't map Windows-style absolute paths (`C:/Users/foo`, `C:\Users\foo`) to the existing domain slug (`c--users-michael-crawford`). It probably handles POSIX paths only.
2. Even when the correct `project=` slug is passed and `hotMemories` populates, the profile body fields (`style`, `entryPoints`, `recurringPatterns`, `blindSpots`) remain empty despite `rebuild_profiles({force: true})` having run successfully and `list_domains` reporting non-zero confidence and `dominantMode: mixed`. So the profile body materialization is partially broken too — possibly a separate issue.

## Suggested fix

For the path normalization side: in `shared/project_ids.py` (path ↔ project ID ↔ label ↔ domain ID conversion), accept Windows-style paths — strip drive letter, replace backslashes with forward slashes before slugification, or detect the existing slug pattern by canonicalizing both sides.

The hollow-profile-body issue likely needs separate investigation in the profile_assembler / profile_builder pipeline.

## Environment

- Cortex 3.14.5, Docker install
- Host: Windows 11
- Calls made from Claude Code on Windows where cwd is `C:/Users/michael.crawford`

cdeust:
Fixed in 4260a99. 

[...]

Issue #17, opened 2026-05-05
FastMCP: remember/recall/get_telemetry return 'structured_content must be a dict' despite success


PSGSupport:
## Summary

Three MCP tools return a FastMCP runtime error to the client even though the underlying memory operation succeeds and persists correctly:

- `mcp__cortex__remember`
- `mcp__cortex__recall`
- `mcp__cortex__get_telemetry`

Error message:

> structured_content must be a dict or None. Got str: '{...}'. Tools should wrap non-dict values based on their output_schema.

The data inside the error payload is the correct JSON; it's just being returned as a stringified JSON instead of a dict, so FastMCP rejects it before delivery.

## Reproducer

From a strict MCP client (e.g. Claude Code's tool runtime), call:

```
remember({content: \"any non-trivial content\", tags: [\"test\"]})
```

The DB write happens (verified by `memory_stats` count incrementing and direct `psql` on the `memories` table), but the client sees the schema-wrap error.

The same shape of error occurs on `recall` and `get_telemetry`. Tools that work cleanly include `memory_stats`, `forget`, `query_methodology`, `list_domains`, `wiki_list`, `wiki_read`, `detect_gaps`, `seed_project`, `consolidate`, `rebuild_profiles`, `backfill_memories`.

## Likely cause

The handlers' return values are being JSON-encoded to a string before FastMCP sees them. FastMCP expects a dict (or None) when an `output_schema` is declared. Either:

1. The handlers wrap the dict in `json.dumps()` somewhere in the response path, or
2. The `output_schema` decorator is missing on these specific handlers and FastMCP's auto-wrap can't infer the structure.

Comparing the affected vs. clean handlers should localize quickly — the cleanly-working handlers presumably return dicts directly without the JSON string round-trip.

## Impact

- Every call from a strict MCP client surfaces as an exception even when the op succeeded — confusing UX, and unsafe for any caller that bails on tool exceptions.
- Workaround: ignore the error and verify via `memory_stats` / direct `recall` from a different code path. Not a production-acceptable workaround.

## Environment

- Cortex 3.14.5
- Python 3.12 inside Docker container (`cortex-mcp:latest` built from upstream source via local Dockerfile)
- FastMCP version: whatever is pinned in `pyproject.toml` (`fastmcp>=2.0.0`)
- Client: Claude Code MCP runtime

cdeust:
[...]

Issue #16, opened 2026-05-05
seed_project: delete_memories_by_tag('seeded') is global, ignores domain scope


PSGSupport:
## Summary

`mcp_server/handlers/seed_project.py:180` purges memories globally by tag, regardless of the `domain` argument:

```python
purged = 0
if not dry_run:
    purged = _get_store().delete_memories_by_tag("seeded")
```

This means every call to `seed_project` wipes all previously-seeded memory rows across every domain — only the most recent seed survives.

## Reproducer

Sequentially seed two repos with distinct domains:

```
seed_project(directory="/path/to/repoA", domain="repo-a")
# stored 4, purged_stale 0

seed_project(directory="/path/to/repoB", domain="repo-b")
# stored 7, purged_stale 4   <-- wiped repoA's rows, despite different domain

memory_stats()
# only repoB's 7 rows + any pre-existing non-"seeded" memories
```

Expected: `purged_stale 0` on the second call, since `repo-b` has no prior seeded memories.
Actual: every prior `seeded`-tagged memory is deleted regardless of domain.

## Impact

- Bootstrapping multiple repos in one session (the natural workflow on a multi-project dev machine) leaves only the last repo's structural-summary / configs / docs / entry-points in memory.
- Knowledge-graph entities and relationships accumulate fine across calls (different table, different lifecycle), so cross-repo recall via the KG still works. The loss is the per-repo discovery memory rows.
- Workaround: end with the repo whose rows you most want to retain.

## Suggested fix

Scope the purge by domain when one is provided:

```
purged = 0
if not dry_run:
    purged = _get_store().delete_memories_by_tag_and_domain("seeded", domain) if domain \
             else _get_store().delete_memories_by_tag("seeded")
```

Or keep the global purge but filter by `directory` (since each seed has a `directory` field on the stored memory).

## Environment

- Cortex 3.14.5, Docker install
- PostgreSQL 15 + pgvector
- Discovered while bootstrapping 8 sibling repos with `seed_project` per repo

## Notes from same session (separate, lower priority)

While verifying the above, also surfaced:

1. `remember`, `recall`, and `get_telemetry` return `structured_content must be a dict or None. Got str: ...` from the FastMCP layer despite the underlying op succeeding. Likely missing/incorrect `output_schema` declaration on those handlers.
2. `query_methodology(cwd="C:/Users/...")` does not slug-normalize Windows-style paths to the existing domain; passing `project=` works.

Happy to file these as separate issues if useful.

[...]

PR #19, opened 2026-05-05
fix(docker): use `python -m mcp_server` entrypoint

PSGSupport:
## Summary

The Dockerfile's `ENTRYPOINT ["neuro-cortex-memory"]` references a console script that is not registered in `pyproject.toml`. The only script declared under `[project.scripts]` is `cortex-doctor`. As a result, the image fails to start:

```
exec: "neuro-cortex-memory": executable file not found in $PATH
```

Switch to `python -m mcp_server`, which is the invocation the package itself documents in `mcp_server/__main__.py`:

```python
"""Bootstrap entry point for the methodology-agent MCP server.

Usage:
    python -m mcp_server
"""
```

## Verification

- `pyproject.toml` `[project.scripts]` only registers `cortex-doctor`
- `mcp_server/__main__.py` is present and exits cleanly via SIGTERM/SIGINT handlers
- Locally built image with this change: container starts, `mcp_server` registers FastMCP tools, MCP stdio handshake succeeds
- HEALTHCHECK still passes (it imports `memory_config`, unrelated to entrypoint)

## Scope

Single-line change inside the existing `Dockerfile` plus a short comment explaining why. No behavioral change beyond making the image actually launch. No changes to dependencies, build stages, or runtime image surface.

Discovered while standing up Cortex on a Windows host via Docker (the only practical install path on Windows since `setup.sh` is gated on `Darwin/Linux` in `detect_os()`). Found in 3.14.5; same line is present in 3.15.0.

## Test plan

- [ ] `docker build -t cortex:test .` succeeds
- [ ] `docker run --rm -e DATABASE_URL=postgresql://... cortex:test` starts (stdio mode; expect to feed/read MCP messages or pipe `/dev/null`)
- [ ] HEALTHCHECK exits 0 within 30s

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,4 +56,8 @@ HEALTHCHECK --interval=30s --timeout=10s --retries=3 \
 
 # MCP servers typically run stdio transport; no ports to expose.
 # Prometheus metrics endpoint is served by the sidecar in Phase 7.1.
-ENTRYPOINT ["neuro-cortex-memory"]
+#
+# Use `python -m mcp_server` because pyproject.toml registers only one
+# console script (`cortex-doctor`); there is no `neuro-cortex-memory`
+# entry point. The package's __main__.py documents this invocation.
+ENTRYPOINT ["python", "-m", "mcp_server"]


cdeust (approved):
Dijkstra correctness check: `pyproject.toml [project.scripts]` only registers `cortex-doctor`; `neuro-cortex-memory` was never installed, so the existing ENTRYPOINT failed at container start. `python -m mcp_server` matches the invocation documented in `mcp_server/__main__.py`.

Diff is 5 lines, the change is non-regressive (current state is broken), HEALTHCHECK is unaffected.

Approving. Thanks for the clean repro + verification — exactly the kind of contributor PR that lands fast.

Will land in v3.15.1 alongside fixes for issues #16, #17, #18 (currently in flight).

cdeust (approved):
Dijkstra correctness check passed: confirmed pyproject.toml only registers `cortex-doctor` (lines 1-9 of [project.scripts]). `neuro-cortex-memory` was never installed, so the previous ENTRYPOINT failed at container start as reported. The new `python -m mcp_server` matches the invocation documented in mcp_server/__main__.py:7.

Diff is 5 lines, non-regressive (current state is broken), HEALTHCHECK unaffected.

Will land in v3.15.1 alongside fixes for #16, #17, #18 (engineers in flight). Thanks for the clean repro.

cdeust:
Clean fix — the `pyproject.toml` evidence + the `mcp_server/__main__.py:7` documented invocation made this a Dijkstra-class correctness fix (the existing entrypoint wasn't reachable; the proposed one is documented). Squash-merged as `50246ed` and shipped in v3.15.1 alongside fixes for the four issues you also reported (#16 Coase / #17 Liskov / #18 Hopper / #20 Feynman). https://github.com/cdeust/Cortex/releases/tag/v3.15.1

PR #48, opened 2026-05-27
ci(pypi): add Trusted Publishing workflow

cdeust:
## Why

`release.yml` only builds a GitHub Release; PyPI uploads have always been by-hand from a maintainer's shell. v3.17.1 (the GHSA-gvpp-v77h-5w8g security fix) is sitting tagged + released on GitHub but unpublished on PyPI because the maintainer's PyPI token isn't visible to the Claude-Code subprocess shell that built it.

## What

A new workflow `.github/workflows/publish-pypi.yml` that uses [PEP 740 Trusted Publishing](https://docs.pypi.org/trusted-publishers/) — GitHub Actions mints a short-lived OIDC token, PyPI verifies it against the trusted-publisher configuration on the project, and issues a one-shot upload token. **No long-lived secret stored anywhere.**

Triggers:
- `push: tags: v*` — every tag auto-publishes
- `workflow_dispatch` — manual re-publish for tags that pre-date this workflow (i.e. `v3.17.1`)

## Setup required on PyPI side (one-time)

After this merges, do this **once** on pypi.org:

1. Log in to pypi.org as a maintainer of `neuro-cortex-memory`.
2. Go to https://pypi.org/manage/project/neuro-cortex-memory/settings/publishing/
3. Click **Add a new publisher** → GitHub.
4. Fill in:
   - Owner: `cdeust`
   - Repository name: `Cortex`
   - Workflow filename: `publish-pypi.yml`
   - Environment name: leave blank (or `pypi` if you also uncomment the `environment:` line in the workflow's publish job)
5. Save.

## After PyPI is configured

I'll re-trigger this workflow against tag `v3.17.1` via `workflow_dispatch` to actually ship the security fix. Subsequent tag pushes will auto-publish without any further action.

## Why now, instead of next release

The security fix for GHSA-gvpp-v77h-5w8g (CVSS 7.8 HIGH) is currently advertised in the published GHSA and GitHub Release but cannot be installed via `pip install --upgrade neuro-cortex-memory`. Every hour that gap stays open is a maintainer-availability risk.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,116 @@
+name: Publish to PyPI
+
+# Trusted Publishing (PEP 740 / OIDC) — no long-lived secret stored in
+# GitHub or anywhere else. The workflow asks GitHub Actions to mint a
+# short-lived OIDC token; PyPI verifies the token against the trusted
+# publisher configuration on https://pypi.org/manage/project/neuro-cortex-
+# memory/settings/publishing/ and issues a one-shot upload credential.
+#
+# To enable this for the first time on PyPI:
+#   1. Log in to pypi.org as a project maintainer.
+#   2. Navigate to the project's "Publishing" settings page.
+#   3. Under "Add a new publisher", choose GitHub and fill in:
+#        Owner:               cdeust
+#        Repository name:     Cortex
+#        Workflow filename:   publish-pypi.yml
+#        Environment name:    (leave blank, or use "pypi" if you want
+#                              an extra approval gate; if you set one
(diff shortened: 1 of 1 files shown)


cdeust:
Superseded: instead of a new `publish-pypi.yml` (which would need a fresh PyPI Trusted Publisher entry), the build+publish-pypi jobs were restored directly into `release.yml` in 9c9b659. That reuses the existing trusted-publisher entry (cdeust/Cortex, release.yml, env=pypi) from the <=3.14.7 era, so no PyPI reconfiguration is needed, and it's wired as a non-blocking deprecated channel per ADR-0050. Closing this PR.

Issue #174, opened 2026-07-24
test: test_rebuild_profiles depends on live ~/.claude data — 3 failures reproduce on main


cdeust:
Dated remediation item (2026-07-25, §13) surfaced during #169: three `test_rebuild_profiles` cases fail identically on main with a `bridge_finder` ValueError sourced from the developer's live `~/.claude` session data. Tests reading the developer's real home directory violate test isolation (results depend on whose machine runs them) — the fixture should be a synthetic projects dir. Classify per test-engineer rules: hidden shared state, not flakiness. Fix: hermetic fixture + regression test for the bridge_finder ValueError itself (it may be a real bug against unusual session data, not just a test problem).

cdeust:
Follow-up merged: PR #183 (cold_start hermetic rewrite). Final state on the developer machine with live ~/.claude data: **5352 passed, 0 failed** (EXITCODE=0, 44m33s) — the first fully green suite of this arc. The I2 ordering failure was adjudicated an environmental confound (a concurrent agent mutating the shared checkout mid-scan; proven empirically via isolated-worktree runs) — no code change warranted, the invariant scan stays strict by design. Test-isolation defect class in this repo: closed.

PR #183, opened 2026-07-25
test(cold-start): hermetic SessionStart banner — kill live-env dependency (#174 follow-up)

cdeust:
Follow-up to #174 / #180 (merged). The first full-suite run after #180 surfaced two more failures in the same test-isolation family; this PR fixes the one that is a genuine test defect.

## `test_cold_start.py::test_normal_session_with_memories` — live-environment dependency

Root cause: the test ran `session_start.py` as a **subprocess under the inherited real environment**. The SessionStart backend gate reads the developer's live backend marker (`~/.claude/methodology/backend.json`); the banner then reads the real `~/.claude/methodology/memory.db` (or a live PostgreSQL); and the subprocess additionally spawned detached consolidate/reanalyze workers against the real HOME and repo. The in-process `remember` seed landed in the per-process test DB while the subprocess read a *different* live store — so the assertion outcome depended on whose machine ran it (failed on this developer's populated home, passed on an empty CI home). The seed's raw `UPDATE ... NOW()` is also PostgreSQL-only, silently mismatched under the SQLite default.

Fix (test-side — the SQLite banner path itself is already proven correct by `tests_py/hooks/test_sqlite_hook_paths.py`): rewritten to the canonical hermetic pattern — seed an ephemeral `SqliteMemoryStore` in `tmp_path` and drive `_sqlite_context` in-process via a monkeypatched `get_shared_store`, asserting the seeded memory is injected and no PostgreSQL install guidance leaks. Deterministic and environment-independent; same behavior verified (DB has memories -> banner injected, not cold start).

## The second failure — `test_I2_no_unauthorized_heat_writes` — is NOT a code defect

`test_I2` is a stateless static source scan (`rglob mcp_server/**/*.py`, compare `SET heat_base` sites to a line-pinned allow-list). It holds no runtime state, so it cannot carry hidden shared state between tests. Its one observed failure occurred in a working tree a **second concurrent agent was actively mutating** (running its own `pytest tests_py/` and editing `test_I2` + `pg_store.py`/`sqlite_store.py` on another branch in the same checkout); a `git checkout`/edit landing mid-run shifts the pinned offsets so the scan reads source that no longer matches the allow-list. It passes standalone and in every controlled/isolated run — no code change is warranted (hardening the scan against a concurrently-mutated tree would only weaken the invariant it guards).

Gates: `ruff check` + `ruff format --check` pass. Full suite verified green in an isolated worktree (own source via PYTHONPATH, own throwaway DB via conftest), immune to the concurrent agent's mutations.

### DIFF
--- a/tests_py/integration/test_cold_start.py
+++ b/tests_py/integration/test_cold_start.py
@@ -41,57 +41,58 @@ def _pg_reachable() -> bool:
 class TestSessionStartHook:
     """Test session_start.py outputs correct context for different DB states."""
 
-    def _run_hook(self, env_overrides: dict | None = None) -> str:
-        """Run session_start.py as subprocess and capture stdout."""
-        hook_path = os.path.join(
-            os.path.dirname(__file__),
-            "..",
-            "..",
-            "mcp_server",
-            "hooks",
-            "session_start.py",
-        )
-        hook_path = os.path.abspath(hook_path)
-        env = {**os.environ, **(env_overrides or {})}
-        result = subprocess.run(
-            [sys.executable, hook_path],
-            capture_output=True,
-            text=True,
-            timeout=15,
-            env=env,
-        )
-        return result.stdout.strip()
+    def test_normal_session_with_memories(self, tmp_path, monkeypatch, capsys):
+        """When the store has memories, the banner injects context (not cold start).
 
-    def test_normal_session_with_memories(self):
-        """When DB has memories, hook should inject memory context."""
-        from mcp_server.handlers.remember import handler as remember
-        import asyncio
+        Hermetic (issue #174 family — hidden live-environment dependency).
+        The prior version ran ``session_start.py`` as a subprocess with the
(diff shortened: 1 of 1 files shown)


cdeust:
## Full-suite proof (this machine, live `~/.claude` present)

Run on current `main` + this cold_start fix, in an isolated git worktree (own source via `PYTHONPATH`, own per-process throwaway DB via `conftest`) so the result is immune to a second agent's concurrent `pytest` + `git checkout` activity in the shared checkout:

```
5352 passed, 12 skipped, 9 warnings in 2673.84s (0:44:33)
EXITCODE=0
```

Zero failures. The three originally-failing `test_rebuild_profiles` cases (fixed in merged #180), the rewritten `test_cold_start::test_normal_session_with_memories`, and `test_I2_no_unauthorized_heat_writes` are all green. I2 needed no code change — it is a stateless static source scan whose one prior failure was concurrent shared-checkout mutation, not hidden state.

Issue #293, opened 2026-07-30
Any two PRs that add tests conflict on six files by construction (test count hard-coded)


cdeust:
## Problem

The collected test count is hard-coded, by hand, in six files: `.bestpractices.json`, `CLAUDE.md`, `CONTRIBUTING.md`, `README.md`, `docs/ASSURANCE-CASE.md`, `assets/badge-tests.svg`. Each is checked for exact equality against whichever branch's own live `pytest --collect-only` count runs in CI.

That count is a property of the post-merge tree, not of any one branch: two branches that each add tests compute two different, both-true numbers and must each edit the same six lines to match. Whichever merges second silently overwrites the first's correct figure with its own now-stale one.

Measured on this repo, not hypothesised:
- 6 of 8 open PRs conflicted on exactly these files; two PRs conflicted on nothing else.
- A PR flipped from MERGEABLE/CLEAN to CONFLICTING/DIRTY the instant a sibling PR merged.
- `main` went red twice between merges purely from this race (PR #280 synced the number to its own total; #278 then added more tests against a now-stale base).
- The same PR was rebased three times purely to resolve this, and four different mechanical resolvers all produced defects.

## Fix (this issue tracks the PR that resolves it)

- `assets/badge-tests.svg` becomes the ONLY artifact stating an absolute test count.
- Its check moves from exact equality to a monotone floor (`committed <= live`) in both `scripts/check_doc_claims.py` (`doc_claim_structural.check_badge_floor`) and `scripts/generate_repo_badges.py` (`stale_tests_badge`) — a committed count that lags the live one is stale-but-true and passes; only an over-claim fails.
- The other five files drop the number entirely, pointing at the badge instead.
- A standing regression test (`RepositoryTests.test_no_prose_file_states_the_suite_size_any_more`) asserts no scanned file — `.bestpractices.json` included — states this claim in prose again.
- `scripts/check_doc_claims.py` (420 lines) and `scripts/generate_repo_badges.py` (305 lines) were both over the repo's 300-line file cap before this work needed to touch them further; split via Extract Module into `doc_claim_sources.py`/`doc_claim_scan.py`/`doc_claim_structural.py`/`repo_badge_catalog.py`, verified behavior-preserving by an unchanged test suite before the floor logic was added.

Acceptance test: two throwaway branches off the fix branch, each adding a trivial passing test, merge with zero conflicts in the six files.

cdeust:
[...]

PR #296, opened 2026-07-30
chore: refresh assets/badge-tests.svg from a measured count (post-#294)

cdeust:
## Summary

Tiny, single-file follow-up to #294 (issue #293). #294's own squash-merge
(05c7169) landed `assets/badge-tests.svg` still stating **6673** — main's
count from before that PR's own rebase — rather than a fresh measurement.
6673 <= 6692 (the true live count on this tree), so the monotone-floor
gate #294 introduced correctly stayed green; the number just isn't
current.

Not a bug in #294's design (a floor is explicitly allowed to lag), but a
floor that is only ever checked, never refreshed, gets stale forever —
worth a one-line regeneration whenever it's touched rather than carried
forward PR after PR.

## Change

`python scripts/generate_repo_badges.py --test-count 6692` (not
hand-edited) — 6692 is `pytest --collect-only -q` on this branch's tip.
No other file changed.

## Completion Ledger (§13.2)

| Path / behavior | Evidence |
|---|---|
| `assets/badge-tests.svg` title/label/alt refreshed to the measured count | Regenerated via the actual generator script, not hand-edited; diffed — only the four count occurrences changed |
| No other doc drifted | `git diff` scoped to this commit touches exactly one file |

## Test plan

- [x] `python scripts/check_doc_claims.py --test-count 6692` -> `doc claims OK (1 declared not-a-claim exemption(s))`
- [x] `python scripts/generate_repo_badges.py --check --test-count 6692` -> `badges OK (5 checked)`
- [x] `pytest tests_py/scripts/test_generate_repo_badges.py -q` -> `49 passed, 20 subtests passed`
- [x] `ruff check scripts/generate_repo_badges.py` -> clean
- [x] Constructed a genuinely false claim (badge overclaiming 999999) against this same tree and confirmed both `check_doc_claims.py` and `generate_repo_badges.py --check` fail closed (exit 1) — the floor gate #294 introduced is still blocking, not just decorative

## Boy-scout check (§14)

- Seen, fixed here: the stale (non-overclaiming) badge count left by #294's landed merge.
- Deferred: none.
- Bypass used: none.


cdeust:
Superseded — no longer needed. Main moved twice more while this was in flight: PR #284 landed right after #294 (main is extremely active right now), and whoever rebased #284 onto main also regenerated `assets/badge-tests.svg` as part of that landing — it now correctly states 6706, which matches a fresh `pytest --collect-only -q` on the current main tip exactly (verified: `check_doc_claims.py --test-count 6706` -> "doc claims OK", `generate_repo_badges.py --check --test-count 6706` -> "badges OK (5 checked)").

This PR's own regeneration (to 6692) was correct against the main tip it was branched from (3f0522b) but would now be a regression (6692 < 6706, reintroducing staleness) against the current tip. Closing rather than rebasing a now-pointless diff.

Issue #298, opened 2026-07-30
Ratchet backlog: 229 file + 607 method size-cap breaches pre-date the new craftsmanship-checker gate (issue #297)


cdeust:
## Context

Issue #297 wires a real, blocking CI gate (`scripts/craftsmanship_checker.py`)
enforcing CLAUDE.md's documented-but-previously-unenforced 300-line-file /
40-line-method (signature-to-close) caps. Running it against the full tree
at commit 2590922 (origin/main, 2026-07-30) found a large pre-existing
backlog, measured via `git ls-files -- '*.py'` + Python `ast`:

```
Scanned 1093 tracked .py files, 12109 functions/methods.
FILE breaches (>300 lines):   229
  by top-level dir: mcp_server 121/527, tests_py 74/459, benchmarks 25/63,
  scripts 8/39, video 1/1, fuzz 0/3, _pipeline 0/1
METHOD breaches (>40 lines, signature-to-close): 607
  by top-level dir: mcp_server 401/3424, benchmarks 74/436, tests_py 85/7937,
  scripts 39/235, video 6/20, fuzz 0/11, _pipeline 0/1
```

Fixing 836 breaches spanning nearly every module in the repo is a
repo-wide refactor an order of magnitude larger than "wire the gate"
(coding-standards.md §15.1: don't mix an unrelated, much-larger root-cause
fix into the gate PR) and would collide with the many other in-flight PRs
already touching these same files.

## What issue #297 does instead
Ships a committed, dated ratchet baseline
(`scripts/craftsmanship_baseline.json`) recording exactly this backlog as
of 2590922. The checker:
- Fails any NEW breach (a file/function not in the baseline going over cap).
- Fails any REGRESSION (a baselined item getting even one line longer).
- Fails a STALE baseline entry (an item fixed or removed must have its
  baseline entry removed in the SAME change that fixes it) — this is what
  makes the ratchet monotonically shrink over time instead of accumulating
  dead entries.
- Passes a baselined item that stays the same or shrinks (still over cap).

## Acceptance criteria for THIS issue (the backlog, not the gate)
- Reduce `scripts/craftsmanship_baseline.json`'s `files` and `methods`
  entries to zero, one PR (or a small number of scoped PRs) at a time,
  each a behavior-preserving Extract-Function/Extract-Module refactor with
  the existing test suite as the proof (passing unchanged) plus new tests
  for any newly-extracted unit, per coding-standards.md §6/§9.
[...]

PR #319, opened 2026-07-31
fix(tests): skip test_phase2_parity on missing schema, not just unreachable PG

### LINKED ISSUE #312: test_phase2_parity.py: PG-reachable-but-wrong-schema causes hard failure instead of skip
## Symptom

`tests_py/invariants/test_phase2_parity.py` (5 tests in `TestPhase2Parity`)
fails hard with `psycopg.errors.UndefinedTable: relation "memory_entities"
does not exist` when run standalone in this environment, instead of the
documented skip ("Uses `cortex_test` PostgreSQL — see `tests_py/conftest.py`.
Skipped when PG is not available").

## Reproduction

```
python -m pytest tests_py/invariants/test_phase2_parity.py -q
# 5 failed, 2 passed (deterministic across repeated runs in this environment)
```

Reproduces identically on unmodified `main` (verified via `git stash` /
`git stash pop` against a clean checkout at the tip of
`fix-lint-remaining-families-239`'s base) — not caused by any in-flight
change. In the full-suite invocation (`pytest tests_py/ -q`) these same
5 tests are reported as **skipped**, not failed — the outcome differs by
invocation context even though `tests_py/conftest.py` sets `DATABASE_URL`
identically in both cases.

## Root cause

The module-level skip gate (`tests_py/invariants/test_phase2_parity.py`
lines ~50-60) only checks bare connectivity:

```python
try:
    conn = psycopg.connect(_PG_URL, autocommit=True, connect_timeout=3)
    conn.close()
    _PG_AVAILABLE = True
except Exception:
    _PG_AVAILABLE = False
```

It does not check that the target database actually has the schema this
test needs (`memory_entities` and friends). When a local/ambient
PostgreSQL instance is reachable at the configured `DATABASE_URL` but its
`cortex_test` database hasn't had the full schema migrated (or points at
a different, unrelated database), the connection succeeds, the skip does
NOT fire, and the test proceeds straight into a hard failure instead of
a clean skip — the same class of defect closed under #219
(`test_consolidate` binding to whatever store the environment ambiently
provides, without verifying it's the right one).

## Why it matters

A flaky pass/fail split driven purely by ambient local infra state (is
some other process's Postgres reachable right now, does it happen to
have the right schema) is not deterministic test isolation
(coding-standards.md §13.1 G3). Contributors and CI runners without a
freshly-migrated `cortex_test` schema get a red suite for a reason
unrelated to their change.

## Suggested fix

Extend the skip-gate query to also confirm the required table(s) exist
(e.g. `SELECT to_regclass('memory_entities')`) and treat "reachable but
wrong/incomplete schema" the same as "unreachable" — skip with a
distinguishing reason, don't fail.

## Related

- #219 (closed) — same category (ambient-store binding without
  isolation), different specific test.

Found while running `tests_py/invariants/test_phase2_parity.py` as part
of the issue #239 lint-remediation program's own verification step
(refactorer agent) — filed per coding-standards.md §14.3 since fixing
the skip-condition logic is a test-infrastructure bug fix, outside a
lint-refactor's blast radius and outside a refactorer's charter (no
behavior change without a dedicated bugfix PR).

cdeust:
Closes #312

## Root cause

`tests_py/invariants/test_phase2_parity.py`'s module-level skip gate checked
only bare PostgreSQL *connectivity* (`psycopg.connect(...).close()`
succeeding), never that the target database has the schema the test body
actually needs (`memories`/`entities`/`memory_entities`/`relationships`). A
reachable-but-unmigrated database — reproduced locally via
`tests_py/_pg_throwaway_db.py`'s per-process throwaway DB, a bare `CREATE
DATABASE` with no tables until some other test in the session happens to
construct a `PgMemoryStore` first — passes the connectivity check and then
dies mid-test with `psycopg.errors.UndefinedTable` instead of skipping. Same
ambient-store-binding defect class as #219, on a different test.

Reproduction (before the fix, standalone run against a schema-less throwaway
DB):
```
python -m pytest tests_py/invariants/test_phase2_parity.py -q
5 failed, 2 passed
```

## Fix

Extracted the schema-presence probe into a standalone, independently unit-
tested helper — `tests_py/invariants/_pg_schema_probe.py` (mirrors the
existing `_pg_safety_guards.py` / `_pg_throwaway_db.py` split rather than
folding more logic into an already 576-line test file). `pg_gate()`:

- Checks every required table via `SELECT to_regclass(%s)` (the issue's own
  suggested approach), never short-circuiting on the first miss so the skip
  reason names every missing table at once.
- Returns `(False, "... not reachable ...")` when PG is unreachable (silent,
  as before — the ordinary "no PostgreSQL here" case).
- Returns `(False, "... missing required table(s) [...] ... issue #312 ...")`
  **and** emits a `SchemaSkipWarning` via `warnings.warn()` when PG is
  reachable but schema is missing — pytest's warnings summary prints
  regardless of `-q` verbosity (unlike a skip *reason*, which needs
  `-rs`/`-ra`), so this distinguishing, surprising condition is visible in
  plain CI output instead of reading as an unrelated, expected skip.
- Returns `(True, "...")` only when every required table resolves.

`test_phase2_parity.py`'s own gate now calls `pg_gate(...)` with the four
tables its fixture/backfill/scan helpers touch, and uses the returned reason
directly in `pytest.mark.skipif`.

**What did NOT change:** the fix only widens the *skip condition*. It does
not touch how a test's own assertions fail. Verified with a deliberately
mutated assertion (`len(pairs) > 999999`) run against the fully migrated
`cortex_test` database — it still fails hard, not silently skips.

## Verification

- Standalone, schema-less throwaway DB (the exact repro): `7 skipped, 1
  warning` — warning text: `SchemaSkipWarning: test_phase2_parity: PostgreSQL
  is reachable at '...' but missing required table(s) ['memories',
  'entities', 'memory_entities', 'relationships'] — schema not migrated
  (issue #312: reachable-but-wrong-schema skips, it does not fail)`.
- Standalone, real migrated `cortex_test`: `7 passed`, no warning.
- Genuinely unreachable PostgreSQL (`CORTEX_TEST_DATABASE_URL` pointed at a
  closed port): `7 skipped`, **no warning** (the ordinary, unsurprising case
  — warning only fires for the surprising reachable-but-wrong-schema case).
- Mutated real-failure control: `1 failed` (assertion still fails hard when
  schema is present and a real bug exists).
- Full suite (`pytest -q`, local throwaway-DB isolation): `7017 passed, 12
  skipped, 1 warning, 121 subtests passed in 178.75s` — the 1 warning is the
  same `SchemaSkipWarning`, now visible where it used to be silent.
- `tests_py/invariants/` alone: `95 passed`.
- New dedicated unit tests for the helper
  (`tests_py/invariants/test_pg_schema_probe.py`, 12 tests): unreachable ->
  false + no warning (incl. `psycopg` absent entirely); missing schema ->
  false + warning naming every missing table, exact reason text, exact
  `stacklevel=2`; full schema -> true + no warning; exact
  `to_regclass(%s)`/`autocommit=True, connect_timeout=3` call pinning.

## Mutation testing (`scripts/mutation_check.sh`, mutmut)

Scoped to the new helper against both new/touched test files:
```
scripts/mutation_check.sh \
  "tests_py/invariants/test_pg_schema_probe.py tests_py/invariants/test_phase2_parity.py" \
  tests_py/invariants/_pg_schema_probe.py
```
Result: **30/30 mutants killed, 0 survivors.** (First pass surfaced 4
survivors — a string-literal wrap/case mutation on the issue-#312 reason
text, and a `stacklevel` argument mutation on `warnings.warn` — both closed
by adding an exact-reason-text assertion and a mocked-`warnings.warn`
stacklevel assertion; see `test_reason_text_is_exact` /
`test_warns_with_stacklevel_2`.)

## Completion Ledger

| # | Path introduced by this diff | Test asserting its observable effect |
|---|---|---|
| 1 | `pg_gate` returns `(False, reason)` when PG unreachable | `test_pg_schema_probe.py::TestPgGateUnreachable::test_returns_false_with_a_not_reachable_reason` |
| 2 | unreachable case emits no warning | `TestPgGateUnreachable::test_emits_no_warning` |
| 3 | `psycopg` import failure treated as unreachable (SQLite-only install) | `TestPgGateUnreachable::test_import_error_is_also_treated_as_unreachable` |
| 4 | `pg_gate` returns `(False, reason)` naming the missing table when schema absent | `TestPgGateMissingSchema::test_returns_false_and_names_the_missing_table` |
| 5 | every required table checked, not short-circuited on first miss | `TestPgGateMissingSchema::test_names_every_missing_table_not_just_the_first` |
| 6 | `SchemaSkipWarning` emitted on missing schema, message == returned reason | `TestPgGateMissingSchema::test_warning_message_matches_the_returned_reason` |
| 7 | exact reason text (issue #312 citation, not a case/wrap-mutated variant) | `TestPgGateMissingSchema::test_reason_text_is_exact` |
| 8 | warning raised with `stacklevel=2` (attributes to the caller's gate line) | `TestPgGateConnectionArgs::test_warns_with_stacklevel_2` (via `TestPgGateMissingSchema` fixture shape) |
| 9 | `pg_gate` returns `(True, reason)` when every table resolves | `TestPgGateFullSchema::test_returns_true_when_every_table_resolves` |
| 10 | full-schema case emits no warning | `TestPgGateFullSchema::test_emits_no_warning` |
| 11 | `psycopg.connect` called with exact `autocommit=True, connect_timeout=3` | `TestPgGateConnectionArgs::test_connects_with_exact_reachability_args` |
| 12 | query is parameter-bound (`SELECT to_regclass(%s)`), not string-interpolated | `TestPgGateConnectionArgs::test_query_is_parameter_bound_not_string_interpolated` |
| 13 | `test_phase2_parity.py`'s own gate now skips (not fails) on schema-less throwaway DB, with visible warning | manual reproduction, quoted above (`7 skipped, 1 warning`) — this is the module-level integration point mutmut's `pytest_add_cli_args_test_selection` also exercises |
| 14 | real assertion failures still fail hard (gate change does not convert failures into skips) | manual mutated-assertion control against migrated `cortex_test`, quoted above (`1 failed`) |

## Pre-push gate

1. Completion Ledger above; every test name copied from the diff via `grep -n "def test_"`, not recalled.
2. AST size check (files created/grown): `_pg_schema_probe.py` 78 lines / max function 39 lines; `test_pg_schema_probe.py` 198 lines / max function 16 lines; `test_phase2_parity.py` net +1 line (576→577, pre-existing over-cap file per the standing #298 waiver — not materially grown).
3. `ruff check .` and `ruff format --check .`: both clean (repo-wide).
4. Full suite: `7017 passed, 12 skipped, 1 warning, 121 subtests passed in 178.75s`.
5. `scripts/check_doc_claims.py`: `doc claims OK`. `scripts/generate_repo_badges.py --check`: `badges OK (4 checked)`.
6. No conflict markers; `.bestpractices.json` parses.

No CHANGELOG entry: this is test-infrastructure only, no consumer-observable
product/contract change (consistent with #219/#276/#287, none of which added
CHANGELOG entries either).

No new issues opened. No follow-up PR needed — the fix is complete within
this diff's blast radius.

### DIFF
--- a/tests_py/invariants/_pg_schema_probe.py
+++ b/tests_py/invariants/_pg_schema_probe.py
@@ -1,0 +1,78 @@
+"""Distinguish "PostgreSQL unreachable" from "PostgreSQL reachable but
+missing the schema a test needs" (issue #312).
+
+`tests_py/invariants/test_phase2_parity.py`'s original module-level skip
+gate checked only bare connectivity (`psycopg.connect(...).close()`
+succeeding). A database that is reachable but has never had the schema
+migrated — e.g. the per-process throwaway database
+`tests_py/_pg_throwaway_db.py` creates for local runs, which is a bare
+`CREATE DATABASE` with no tables until some other test in the same session
+happens to construct a `PgMemoryStore` first — passes that check and then
+dies mid-test with `psycopg.errors.UndefinedTable` instead of skipping,
+exactly the class of defect closed for a different test under #219
+(ambient-store binding without verifying the store is the right one).
+
+`pg_gate()` below extends the connectivity probe with a schema-presence
+check (`SELECT to_regclass(<table>)`, per the issue's own suggested fix)
+so "reachable but wrong schema" is treated the same as "unreachable": skip,
+with a distinguishing reason, never a hard failure. It does NOT touch how
+a test's own assertions fail — a genuine assertion failure inside a test
+body still fails; only the module-level connectivity/schema gate changes.
+"""
+
+from __future__ import annotations
+
+import warnings
+
+
+class SchemaSkipWarning(UserWarning):
+    """A test module skipped because PostgreSQL is reachable but missing
+    required schema.
+
+    Raised via `warnings.warn`, not as an exception: pytest prints a test's
+    skip reason only with `-rs`/`-ra` (CI here runs plain `-q`), but always
+    prints the warnings summary regardless of verbosity — so a permanently
+    -skipping suite surfaces this in CI output instead of silently reading
+    as an unrelated, expected "no PostgreSQL" skip.
+    """
+
+
+def pg_gate(
+    pg_url: str,
+    required_tables: tuple[str, ...],
+    *,
+    label: str,
+) -> tuple[bool, str]:
+    """Return `(available, skip_reason)` for a module-level PG skip gate.
+
+    `available` is True only when `pg_url` is reachable AND every name in
+    `required_tables` resolves via `to_regclass` — bare reachability is
+    not sufficient (issue #312).
+
+    Every table in `required_tables` is checked (not short-circuited on
+    the first miss) so the reason names every missing table at once,
+    rather than requiring repeated runs to discover them one at a time.
+    """
+    try:
+        import psycopg
+
+        with psycopg.connect(pg_url, autocommit=True, connect_timeout=3) as conn:
+            missing = [
+                table
+                for table in required_tables
+                if conn.execute("SELECT to_regclass(%s)", (table,)).fetchone()[0]
+                is None
+            ]
+    except Exception:
+        return False, f"{label}: PostgreSQL is not reachable at {pg_url!r}"
+
+    if missing:
+        reason = (
+            f"{label}: PostgreSQL is reachable at {pg_url!r} but missing "
+            f"required table(s) {missing!r} — schema not migrated "
+            "(issue #312: reachable-but-wrong-schema skips, it does not fail)"
+        )
+        warnings.warn(SchemaSkipWarning(reason), stacklevel=2)
+        return False, reason
+
+    return True, f"{label}: PostgreSQL reachable with required schema present"
--- a/tests_py/invariants/test_pg_schema_probe.py
+++ b/tests_py/invariants/test_pg_schema_probe.py
@@ -1,0 +1,238 @@
+"""Tests for tests_py/invariants/_pg_schema_probe.py (issue #312).
+
+Each test traces to one clause of the probe's contract:
+  - unreachable PostgreSQL -> (False, reason naming "not reachable"), no
+    warning (that is the ordinary, expected "no PG here" case — warning on
+    it would be noise on every SQLite-only run).
+  - reachable but missing required table(s) -> (False, reason naming every
+    missing table + issue #312), AND a `SchemaSkipWarning` so the condition
+    is visible in CI's warnings summary under plain `pytest -q` (which does
+    not print skip reasons without `-rs`/`-ra`).
+  - reachable with full schema -> (True, reason), no warning.
+  - every required table is checked, not just the first miss (a fixture
+    with two missing tables must name both).
+  - the query is parameter-bound (`%s`, not string-interpolated) and uses
+    the exact `to_regclass` shape — pins the SQL text a mutant could
+    otherwise weaken into an injection-prone f-string undetected by any
+    other test in this file.
+
+Mocking mirrors `tests_py/invariants/test_pg_throwaway_db.py`: a fake
+`psycopg` module is injected via `mock.patch.dict(sys.modules, ...)` so no
+real network/DB call is made, and `conn.__enter__`/`__exit__` are stubbed
+because `pg_gate` uses `with psycopg.connect(...) as conn:`.
+"""
+
+from __future__ import annotations
+
+import sys
+import types
+import unittest.mock as mock
+import warnings
+
+import pytest
+
+from tests_py.invariants._pg_schema_probe import SchemaSkipWarning, pg_gate
+
+
+def _fake_conn(table_oids: dict[str, object]) -> mock.MagicMock:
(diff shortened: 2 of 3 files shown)


cdeust:
Superseded by #320, which PROVISIONS the test schema instead of skipping.

This PR's skip guard was implemented to my (the coordinator's) instruction, and that instruction was wrong. The maintainer identified why: the schema was present before — these tests ran against the real `cortex` database until the #219 store-isolation work redirected them to `cortex_test`, which nothing ever provisions. So the absence is a regression to repair, not a condition to tolerate.

Skipping would have taken a local run from ~7000 assertions to a silent pass and enshrined a local-vs-CI divergence — the same defect class as #287, and the fourth 'gate that reports success without checking' found this week.

#320 keeps the one genuinely correct idea from this PR: the warning for the case that still legitimately skips (PostgreSQL genuinely absent), retargeted as PostgresUnavailableWarning.

PR #320, opened 2026-07-31
fix(tests): provision the test-database schema instead of skipping (issue #312)

### LINKED ISSUE #312: test_phase2_parity.py: PG-reachable-but-wrong-schema causes hard failure instead of skip

cdeust:
Closes #312

Supersedes #319 — please close #319 without merging. That PR detected "reachable but wrong schema" and turned it into a skip; this PR closes the actual gap instead: it **provisions** the schema, so the tests run and assert for real. Its `SchemaSkipWarning` idea is carried over below, retargeted at the one condition that still legitimately skips.

## Root cause (unchanged from the issue, confirmed)

`tests_py/conftest.py`'s per-process throwaway database (`tests_py/_pg_throwaway_db.py`) is a bare `CREATE DATABASE` — zero tables. Before #219, these tests ran against the real `cortex` database, which already carried a full schema; isolation redirected them to a throwaway target without ever provisioning it. CI's shared, persistent `cortex` database hides the gap because *some* earlier test in that session happens to construct a `PgMemoryStore()` first and incidentally provisions the schema for everyone after it — accidental, execution-order-dependent, not deterministic isolation. A standalone or narrowly-scoped local run has no such earlier test, so the first raw-SQL query against `memory_entities` dies with `psycopg.errors.UndefinedTable`.

## Fix

`tests_py/conftest.py`'s `_pg_available()` now delegates to a new module, `tests_py/_pg_schema_provision.py::ensure_pg_ready(url)`, which runs **once per session**, before any PG-gated test executes:

1. Direct-connect succeeds (the common case: the throwaway DB was just created) → apply schema, done.
2. Direct-connect fails but a maintenance connection to `postgres` succeeds (server up, only the named database is missing) → `CREATE DATABASE`, then apply schema.
3. Neither connects (genuinely absent PostgreSQL — the ordinary SQLite-only-install case) → return `False` (unchanged skip outcome), now paired with **one** `PostgresUnavailableWarning` per session so a session-wide skip is visible in the warnings summary instead of requiring `-rs`/`-ra` (PR #319's `SchemaSkipWarning` idea, carried over and retargeted — the "reachable but wrong schema" condition it warned about no longer exists as a code path at all; it's provisioned, not detected).
4. A database that cannot be created (no `CREATEDB` privilege) or schema that cannot be applied (e.g. missing extension) is a **real environment error** and aborts the session via `pytest.exit(..., returncode=2)` with an actionable message naming the URL and the fix — the same fail-closed idiom `tests_py/_pg_safety_guards.py`'s two existing guards already use. It is never downgraded to a skip.

**Reuses the production DDL path — no hand-copied schema.** `_provision_schema` constructs `PgMemoryStore(database_url=url)`, exactly what `mcp_server/migrate.py` does to migrate a live install ("constructing `PgMemoryStore` IS the migration" — that module's own docstring). `PgMemoryStore.__init__` → `_init_schema()` → `pg_schema.get_all_ddl()`, hash-gated so a database already at the code's revision costs one indexed `SELECT`.

`tests_py/invariants/test_phase2_parity.py` (the test named in the issue) drops its own ad-hoc bare-connectivity probe and imports the shared `_USE_PG`/`_TEST_DB_URL` from conftest — the same convention 14 other PG-only test modules already use (`test_pg_pool.py`, `test_pg_supersession.py`, `test_migrate.py`, …). This closes the gap for all of them, not just the one named test, since they all shared the exact same accidental-provisioning exposure.

## Proof — the exact reported repro, now passing

```
$ python -m pytest tests_py/invariants/test_phase2_parity.py -q
.......                                                                  [100%]
7 passed in 0.52s
```
Reproduced 4/4 across repeated standalone runs (deterministic, not order-dependent). Before this fix, the same command on unmodified `main` reproduces the issue's own reported `5 failed, 2 passed`.

Manual, from-scratch verification against a real bare throwaway database (before calling the fix):
```
psycopg.errors.UndefinedTable: relation "memory_entities" does not exist
```
— confirmed the repro is real, then confirmed `ensure_pg_ready()` brings that same database to `to_regclass('memories')`/`'entities'`/`'memory_entities'`/`'relationships'` all non-null, with no further UndefinedTable.

## Completion Ledger

| # | Path introduced by this diff | Test asserting its observable effect |
|---|---|---|
| 1 | `_can_connect` True on successful connect | `test_pg_schema_provision.py::TestCanConnect::test_returns_true_when_connect_succeeds` |
| 2 | `_can_connect` False on connect exception | `TestCanConnect::test_returns_false_when_connect_raises` |
| 3 | `_can_connect` False when psycopg absent | `TestCanConnect::test_returns_false_when_psycopg_is_not_installed` |
| 4 | `_can_connect` exact reachability args (`autocommit=True, connect_timeout=3`) | `TestCanConnect::test_connects_with_exact_reachability_args` |
| 5 | `_database_name` extracts name after rightmost `/` | `TestDatabaseName::test_extracts_name_after_rightmost_slash` |
| 6 | `_database_name` drops query string first | `TestDatabaseName::test_drops_query_string_first` |
| 7 | `_database_name` handles a query string containing its own `/` | `TestDatabaseName::test_query_string_containing_its_own_slash_does_not_confuse_it` |
| 8 | `_create_database` no-ops when DB already exists | `TestCreateDatabase::test_returns_early_when_database_already_exists` |
| 9 | `_create_database` creates DB when missing, exact SQL | `TestCreateDatabase::test_creates_database_when_missing` |
| 10 | `_create_database` connects to maintenance URL, not the target | `TestCreateDatabase::test_connects_to_the_maintenance_url_not_the_target` |
| 11 | existence check is parameter-bound (`%s`, not interpolated) | `TestCreateDatabase::test_existence_check_is_parameter_bound` |
| 12 | concurrent-create race (`DuplicateDatabase`) swallowed, not escalated | `TestCreateDatabase::test_duplicate_database_race_is_swallowed_not_escalated` |
| 13 | other create failure escalates via `pytest.exit`, exact actionable message | `TestCreateDatabase::test_other_create_failure_calls_pytest_exit_with_actionable_message` |
| 14 | `_provision_schema` constructs `PgMemoryStore(database_url=url)` and closes it | `test_pg_schema_provision_ready.py::TestProvisionSchema::test_constructs_store_with_the_given_url_and_closes_it` |
| 15 | construction failure escalates via `pytest.exit`, exact actionable message | `TestProvisionSchema::test_construction_failure_calls_pytest_exit_with_actionable_message` |
| 16 | no `UnboundLocalError` under a mocked (non-raising) `pytest.exit` | `TestProvisionSchema::test_construction_failure_never_raises_unboundlocalerror` |
| 17 | `ensure_pg_ready` direct-success path: exact `_can_connect(url)` arg, provisions, skips DB creation | `TestEnsurePgReady::test_direct_connect_success_provisions_schema_without_creating_db` |
| 18 | genuinely-unreachable path: False, exactly one warning, exact message | `TestEnsurePgReady::test_genuinely_unreachable_returns_false_and_warns_once` |
| 19 | that warning's `stacklevel=2` | `TestEnsurePgReady::test_unreachable_warning_uses_stacklevel_2` |
| 20 | maintenance-reachable path: create then provision, no warning | `TestEnsurePgReady::test_maintenance_reachable_creates_then_provisions` |
| 21 | only the genuinely-absent branch ever warns | `TestEnsurePgReady::test_only_the_genuinely_absent_branch_ever_warns` |
| 22 | `PostgresUnavailableWarning` is a `UserWarning` | `test_postgres_unavailable_warning_is_a_user_warning` |
| 23 | bare throwaway DB genuinely reproduces `UndefinedTable` (pre-fix sanity) | `test_pg_schema_provision_live.py::TestBareThrowawayDatabaseGetsProvisioned::test_bare_database_reproduces_the_reported_defect_unfixed` |
| 24 | `ensure_pg_ready` provisions every required table on a bare DB | `...::test_ensure_pg_ready_provisions_every_required_table` |
| 25 | provisioned DB no longer raises `UndefinedTable` | `...::test_a_provisioned_database_no_longer_raises_undefined_table` |
| 26 | provisioning is idempotent on a second call | `...::test_provisioning_is_idempotent_on_a_second_call` |
| 27 | a genuinely-missing database gets created + provisioned end to end | `TestMissingDatabaseGetsCreated::test_ensure_pg_ready_creates_and_provisions_a_missing_database` |
| 28 | a real unreachable server returns False + warns once | `TestGenuinelyUnreachableStillDegradesGracefully::test_a_real_unreachable_server_returns_false_and_warns_once` |
| 29 | a real `NOCREATEDB` role triggers `pytest.exit`, not a skip | `TestPermissionDeniedFailsLoudlyNotSilently::test_a_role_without_createdb_triggers_pytest_exit_not_a_skip` |
| 30 | `test_phase2_parity.py` now shares the session `_USE_PG` flag | the file's own 7 tests, standalone repro quoted above |

## Mutation testing (`scripts/mutation_check.sh`, mutmut)

Scoped to `tests_py/_pg_schema_provision.py` against all three of its test files:
```
scripts/mutation_check.sh \
  "tests_py/invariants/test_pg_schema_provision.py tests_py/invariants/test_pg_schema_provision_ready.py tests_py/invariants/test_pg_schema_provision_live.py" \
  tests_py/_pg_schema_provision.py
```
Result: **85 mutants, 84 killed, 0 unexplained survivors.** First pass surfaced 19 survivors — all string-literal case/wrap mutations on the two `pytest.exit`/`warnings.warn` message texts (substring assertions don't pin the exact text) plus two `stacklevel`/call-argument gaps — closed by adding exact-text and exact-argument assertions (`test_construction_failure_calls_pytest_exit_with_actionable_message`, `test_genuinely_unreachable_returns_false_and_warns_once`, `test_unreachable_warning_uses_stacklevel_2`, `test_direct_connect_success_provisions_schema_without_creating_db`). The 85th mutant (`_can_connect`'s exception-handler `return False`→`return True`) mutmut itself reported "interrupted by user" (a tool-timing artifact, not a real signal); manually re-applying that exact one-line mutation and running the three test files directly confirms it is killed in 0.28s (`1 failed, 1 passed`, `TestCanConnect::test_returns_false_when_connect_raises`) — 85/85 effectively verified.

## Isolation audit (Move 2)

Every throwaway database and role the new live-PG tests create is scoped to a single test/fixture and dropped in a `finally`/fixture teardown; none touch `_TEST_DB_URL` (the session's own database). Verified no leftover `cortex_test_pw*`/`cortex_312_*` databases or roles after repeated runs.

## What did NOT change — real failures still fail

Deliberately mutated `test_phase2_parity.py::test_substring_scan_returns_nonempty_set`'s assertion to `len(pairs) > 999999` and reran: `1 failed, 6 passed` — a genuine defect still fails hard, never silently skips. Reverted before commit.

## Pre-push gate (this exact head)

- `ruff check .` / `ruff format --check .`: clean, repo-wide.
- Full suite, PostgreSQL backend, this exact head: **7071 passed, 5 skipped, 121 subtests passed** (the 5 skips are `sqlite-vec not installed` — unrelated to PostgreSQL/schema, matching the PostgreSQL CI job's own skip count exactly; **zero** schema-related skips remain, versus this issue's own baseline evidence of 5 PG-job skips before the fix).
- AST size check (new/grown files, measured not asserted): `_pg_schema_provision.py` 208 lines / longest fn 34; `test_pg_schema_provision.py` 195/22; `test_pg_schema_provision_ready.py` 174/22; `test_pg_schema_provision_live.py` 215/39; `conftest.py` grown to 298/35 (under the 300/40 caps). `test_phase2_parity.py` sits at 568 lines / `_setup_fixture` 55 — **pre-existing** (576→568, net **reduced** by this diff) — covered by the standing waiver (#298 closed; CLAUDE.md:74-79).
- `scripts/check_doc_claims.py` / `scripts/generate_repo_badges.py --check`: both OK (also re-verified with `--test-count 7076`, this branch's live `--collect-only` count — the committed tests badge is a documented *floor*, not an exact match; 6788 ≤ 7076 stays valid, so no badge regeneration needed).
- No conflict markers; `.bestpractices.json` parses.
- No CHANGELOG entry: test-infrastructure only, no consumer-observable contract change — same precedent as #219/#276/#287/#319 (none added one either).

## Finding outside this PR's blast radius (disclosed, not fixed here)

While running the full suite repeatedly to verify the above, `tests_py/handlers/consolidation/test_near_dup_calibration_pass.py::TestRunNearDupSample::test_sample_includes_seeded_pair_with_contents` failed **once**, in one of five total full-suite runs across this session. This file is untouched by this diff. I ran a paired-control experiment before concluding anything (coding-standards.md §6.1 — reproduce before you fix, never blame blind):

- Clean `origin/main`, no changes: 7042 passed, 0 failed.
- This diff minus the 3 new test files (conftest.py + test_phase2_parity.py only): 7042 passed, 0 failed — identical count to clean main.
- This diff, complete, run twice: once 7070 passed/0 failed, once **1 failed**/7070 passed, once (after adding the mutation-killing assertions) 7071 passed/0 failed.

The exact same diff produced both a pass and a fail across separate runs, and the reduced diff (conftest.py + test_phase2_parity.py, no new files) never reproduced it — so this is not deterministically caused by this PR. It reproduces standalone every time (`5 passed`), so it is order/session-state dependent. Best-effort, *unconfirmed* hypothesis: `pg_store_near_dup.py`'s anchor CTE (`SELECT id, embedding FROM current_memories ... LIMIT %(anchor_limit)s`) has no `ORDER BY` before its `LIMIT`, and the module's own docstring already documents its HNSW top-K scan as an approximation with no `ef_search` tuning — a plausible but *not verified* mechanism. I am not opening a second PR for it: I don't have a confirmed root cause, and shipping an unverified change to an unrelated subsystem (near-duplicate detection) would itself violate root-cause discipline (§6.1 — fix at the classified source, not a guess). Flagging with full reproduction evidence here rather than silently omitting it, per this repo's own boy-scout standard.

### DIFF
--- a/tests_py/_pg_schema_provision.py
+++ b/tests_py/_pg_schema_provision.py
@@ -1,0 +1,207 @@
+"""Ensure the resolved test-database URL is reachable AND schema-provisioned
+before any PostgreSQL-gated test in this session runs (issue #312).
+
+ROOT CAUSE: `tests_py/_pg_throwaway_db.py`'s per-process throwaway database
+(and, equally, an explicit `CORTEX_TEST_DATABASE_URL` pointed at a database
+that was never migrated) is a bare `CREATE DATABASE` — zero tables. Before
+the #219 store-isolation work, these tests ran against the real `cortex`
+database, which already carried a full schema; isolation redirected them to
+a throwaway target without ever provisioning it. In CI's shared, persistent
+`cortex` database, the gap is invisible: SOME earlier test in the same
+session constructs a `PgMemoryStore()` first and incidentally provisions
+the schema for everyone after it. Standalone or narrowly-scoped local runs
+have no such earlier test, so the very first raw-SQL query against e.g.
+`memory_entities` dies with `psycopg.errors.UndefinedTable` — accidental,
+execution-order-dependent provisioning, not deterministic isolation.
+
+PR #319 proposed detecting this condition and skipping. That is the wrong
+fix (issue #312, maintainer directive): a skip converts thousands of real
+assertions into a silent pass and permanently enshrines the local-vs-CI
+divergence. This module makes provisioning DETERMINISTIC instead — every
+session ensures its own resolved database is ready, exactly the same way
+CI accidentally ends up ready, without depending on which test happens to
+run first.
+
+Reuses the production migration path — never a hand-copied schema.
+Constructing `PgMemoryStore(database_url=...)` under the hood runs
+`_init_schema()`, which applies `pg_schema.get_all_ddl()` idempotently
+(hash-gated: a DB already at the code's revision costs one indexed
+SELECT). This is the exact same call `mcp_server/migrate.py` makes to
+migrate a live install — "constructing PgMemoryStore IS the migration"
+(that module's own docstring) — so a test session and a real deployment
+can never drift onto two different schemas.
+
+Genuinely-absent PostgreSQL (driver not installed, or the server itself
+unreachable) still degrades to a skip, exactly as documented before this
+change. Reachable-but-unprovisioned (the actual #312 defect: server up,
+named database missing or empty) no longer skips — it gets created and/or
+migrated. A REAL environment error while doing so (no CREATEDB privilege,
+a required extension the connecting role cannot install) is reported
+loudly via `pytest.exit` — the same fail-closed idiom already used by this
+file's siblings, `guard_against_populated_db` /
+`guard_against_real_data_roots` in `tests_py/_pg_safety_guards.py` — never
+downgraded to a silent skip; that downgrade is exactly the class of bug
+this module exists to close.
+
+`PostgresUnavailableWarning` carries over PR #319's `SchemaSkipWarning`
+idea (a `UserWarning` subclass raised via `warnings.warn`, printed in
+pytest's warnings summary regardless of `-q` verbosity, unlike a skip
+*reason* which needs `-rs`/`-ra`) retargeted at the one skip-worthy
+condition that survives this change: genuine unavailability. It fires
+exactly once per session, here, rather than once per PG-gated test module
+— fourteen-plus files in this suite skip on the shared `_USE_PG` flag
+(`tests_py/conftest.py`), so a per-module warning would be fourteen-plus
+near-duplicate lines for one underlying condition.
+"""
+
+from __future__ import annotations
+
+import warnings
+
+import pytest
+
+from tests_py._pg_throwaway_db import maintenance_url
+
+
+class PostgresUnavailableWarning(UserWarning):
+    """PostgreSQL is not reachable at all this session (server down, or
+    the `psycopg` driver itself is not installed — the ordinary,
+    expected state of a SQLite-only install). Every PG-gated test skips,
+    as documented; this warning only makes that fact visible in the
+    warnings summary instead of requiring `-rs`/`-ra` to see the reason.
+    """
+
+
+def _can_connect(url: str) -> bool:
+    """True iff a direct connection to `url` succeeds.
+
+    Every failure mode — driver absent, server down, auth rejected, the
+    target database not existing — reads as False here; distinguishing
+    WHY is `ensure_pg_ready`'s job (it tries a maintenance connection
+    next), not this probe's.
+    """
+    try:
+        import psycopg  # noqa: PLC0415 — deferred: optional [postgresql] extra, absent on SQLite-only installs
+
+        with psycopg.connect(url, autocommit=True, connect_timeout=3):
+            pass
+        return True
+    except Exception:  # noqa: BLE001 — reachability probe; every failure reads as "not reachable now" (mirrors tests_py/conftest.py's own _pg_available)
+        return False
+
+
+def _database_name(url: str) -> str:
+    """The database-name segment of a ``postgresql://`` URL.
+
+    Same parsing idiom as `_pg_throwaway_db.maintenance_url`: the query
+    string is dropped FIRST (``partition("?")``), then the name is
+    everything after the rightmost ``/`` in what remains
+    (``rpartition("/")``). Order matters: a query string carrying its
+    own ``/`` (e.g. a file-path option) would otherwise be mistaken for
+    the database-name separator if the rightmost-``/`` split ran first —
+    see that module's test for the exact scenario this ordering avoids.
+    """
+    base, _, _query = url.partition("?")
+    _prefix, _, name = base.rpartition("/")
+    return name
+
+
+def _create_database(url: str) -> None:
+    """Create the database named in `url` via a maintenance connection.
+
+    Pre: the PostgreSQL SERVER is reachable (the caller has already
+    confirmed this by connecting to its maintenance database) — `url`'s
+    own database is the only thing missing.
+    Post: the database exists. A database that cannot be created (no
+    CREATEDB privilege, quota, any other server-side refusal) aborts the
+    session via `pytest.exit`, naming the URL and the concrete fix —
+    never a silent skip (issue #312).
+    """
+    import psycopg  # noqa: PLC0415 — deferred: optional [postgresql] extra, absent on SQLite-only installs
+
+    name = _database_name(url)
+    maint = maintenance_url(url)
+    try:
+        with psycopg.connect(maint, autocommit=True, connect_timeout=3) as conn:
+            exists = conn.execute(
+                "SELECT 1 FROM pg_database WHERE datname = %s", (name,)
+            ).fetchone()
+            if exists:
+                return
+            conn.execute(f'CREATE DATABASE "{name}"')
+    except psycopg.errors.DuplicateDatabase:
+        return  # a concurrent process created it between our SELECT and CREATE
+    except Exception as exc:  # noqa: BLE001 — last-resort boundary: reported via pytest.exit below, never swallowed
+        pytest.exit(
+            f"REFUSING to run: PostgreSQL is reachable at {maint!r} but "
+            f"database {name!r} does not exist and could not be created: "
+            f"{exc}. Fix: grant CREATEDB to the connecting role "
+            f"(ALTER ROLE <role> CREATEDB;) or create it manually "
+            f"(createdb {name}), then re-run (issue #312).",
+            returncode=2,
+        )
+
+
+def _provision_schema(url: str) -> None:
+    """Apply Cortex's production DDL to `url` via the exact path every
+    real construction already uses — never a second, hand-copied
+    definition of the schema (issue #312's explicit requirement; see
+    `mcp_server/migrate.py`, which does the identical thing to migrate a
+    live install: "constructing PgMemoryStore IS the migration").
+
+    Idempotent: `PgMemoryStore._init_schema` hash-gates its DDL apply, so
+    a database already at the code's revision costs one indexed SELECT
+    — this is cheap to call on every session, not just the first.
+    """
+    from mcp_server.infrastructure.pg_store import (  # noqa: PLC0415 — deferred: hard-imports pgvector/psycopg/psycopg_pool, absent on SQLite-only installs
+        PgMemoryStore,
+    )
+
+    try:
+        store = PgMemoryStore(database_url=url)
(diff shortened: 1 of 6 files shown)


cdeust:
CI caught a real gap in the first push: **Test (SQLite backend)** failed —
`TestProvisionSchema`'s `mock.patch("mcp_server.infrastructure.pg_store.PgMemoryStore", ...)`
is string-targeted, so `mock.patch` must import `pg_store` to resolve the
attribute, and that module hard-imports `psycopg`/`pgvector`/`psycopg_pool`
at its own top level unconditionally — absent on a SQLite-only install (no
`[postgresql]` extra). Every other mocked test in this PR patches at the
`provision`-module level instead (never triggers a real `pg_store` import),
so only this one class was exposed.

Reproduced locally against a disposable venv built from
`requirements/ci-sqlite.txt` (matching CI exactly) before fixing, to avoid
guessing: `ModuleNotFoundError: No module named 'psycopg'` inside
`pg_store.py`, surfaced by `mock.patch`'s string-resolution as
`AttributeError: module 'mcp_server.infrastructure' has no attribute
'pg_store'`.

Fix (`e638d8c`): scoped `@requires_psycopg` (the marker `tests_py/conftest.py`
already exports and 3 other test files already use for exactly this class of
gap) onto `TestProvisionSchema` only — `TestEnsurePgReady` and the standalone
warning test keep running on SQLite-only installs since they never need a
real `pg_store` import. Verified against both environments locally before
re-pushing: SQLite-only venv `19 passed, 4 skipped` (3 newly-skipped
`TestProvisionSchema` tests + the pre-existing `test_pg_schema_provision_live.py`
skip, matching the expected psycopg-absent set), PostgreSQL-extra venv
`29 passed` unchanged.

Issue #363, opened 2026-08-06
session_start injects memories verbatim — no applicability critique, and MemHarness measures verbatim replay as actively harmful (76.4% → 70.1%)


cdeust:
MemHarness ([arXiv 2607.28272](https://arxiv.org/abs/2607.28272), 2026-07-30) reports the result that matters most for Cortex's injection hooks: **injecting retrieved memory unmodified is worse than injecting nothing at all.**

## The measurement

Table 2, ALFWorld:

| Arm | ALFWorld | WebShop |
|---|---|---|
| RL only, no memory | 76.4% | 66.1% |
| RL + raw memory (verbatim, no reconstruction) | **70.1%** | 72.6% |
| MemHarness (critique + reconstruct) | **85.2%** | 75.6% |

Their conclusion, verbatim: *"Injecting unrefined memory… hurts ALFWorld performance. This proves verbatim replay introduces state-mismatch noise, which our reconstruction mechanism effectively mitigates."* Raw memory cost 6.3 points against the no-memory baseline. They call this **negative transfer**: *"A memory may be semantically relevant to the task yet inappropriate for the current interaction state, because it was formed under different environmental conditions."*

Counterfactual probe (Table 4a): rejection rates rise from 8.7% to 13.3% when the source state is randomised — the critique step is genuinely reading state, not pattern-matching relevance.

## Current state (verified 2026-08-06)

`mcp_server/hooks/session_start.py` injects anchored + hot memories directly into the context block (see its module docstring, and `_fetch_anchors` at line 130), stamped with injection receipts via `handlers/injection_receipts.py`. Nothing between selection and injection compares the memory's formation context to the current one.

The nearest existing mechanism is **not** equivalent: `core/abstention_gate.py` is a fine-tuned DistilBERT trained on BEAM (query, passage, relevant/irrelevant) pairs — its own docstring says it "filters retrieval results that don't actually answer the query." That is a **relevance** gate. MemHarness's critique is an **applicability** gate: same relevance, different state. A memory can pass BEAM relevance and still be exactly the negative-transfer case. `core/memory_rules.py` filtering and `core/reranker.py` scoring are likewise relevance-shaped.

Note the asymmetry in their own numbers — raw memory *helped* WebShop (+6.5) while *hurting* ALFWorld (−6.3). So the effect is task-dependent and we must measure it on our own workload rather than assume the direction.

## Ask

[...]

Issue #98, opened 2026-07-12
Registered tool signatures drift from handler schemas: remember lacks supersedes_id/write_class, wiki_write lacks memory_ids — promotion loop cannot be closed over MCP


mbe14:
**Environment:** Cortex plugin 4.13.1, Windows 11, Claude Code as MCP client (validates tool calls against the declared input schema; `additionalProperties` violations are rejected client-side).

## Summary

The client-visible input schemas for `remember` and `wiki_write` are derived by FastMCP from the registered wrapper signatures — and those signatures have drifted from the handlers' own schemas and from the instructions your grooming planners emit. Net effect: **the documented lesson-promotion loop cannot be closed over MCP**, and `get_grooming_health` reports `promotion: last_run=null, stale=true` forever, even after every candidate has been processed.

## The drift, file by file

**1. `remember` — missing `supersedes_id` and `write_class`**

- Registered signature (`mcp_server/tool_registry_memory.py:93-101`, and the rooted variant at `:65-72`):

  ```python
  async def tool_remember(
      content: str,
      tags: list[str] | None = None,
      directory: str | None = None,
      domain: str | None = None,
      source: str | None = None,
      force: bool = False,
      agent_topic: str | None = None,
  ) -> dict:
  ```

  No `supersedes_id`, no `write_class`. Your own comment at `tool_registry_memory.py:58-61` states that FastMCP derives the input schema from this signature — so clients never see these params.

- Internal handler schema **does** declare both: `mcp_server/handlers/remember_schema.py:182` (`write_class`, enum auto/deliberate/derived/mechanical) and `:226` (`supersedes_id`).

- Even if a permissive client passed them, the wrapper forwards a hand-built dict (`tool_registry_memory.py:103-115`) containing only content/tags/directory/domain/source/force/agent_topic — **the params are silently dropped before the handler**.

- Meanwhile `lesson_promotion`'s instructions tell the in-session LLM to call `remember(content=..., supersedes_id=<memory_id>, tags=[..., 'promoted:<kind>'], write_class='deliberate', force=true)` to close the loop, and `curate_distill`'s job prompts require `write_class='deliberate'`. Neither call is possible.

**2. `wiki_write` — missing `memory_ids`**

- Registered signature (`mcp_server/tool_registry_wiki.py:58-63`): `path, content, mode, tags`. The forwarded dict (`:67-72`) also omits `memory_ids`.
[...]

PR #375, opened 2026-08-06
fix(capture): decide capture and gate bypass by channel, not by content (#365)

### LINKED ISSUE #365: security: WebFetch/WebSearch output is auto-captured into long-term memory on a keyword the content controls — end-to-end poisoning path
A 2026 survey on long-term memory security ([arXiv 2604.16548](https://arxiv.org/abs/2604.16548)) taxonomizes memory-lifecycle attacks and names "environment-injected poisoning (web pages, tool outputs as contamination vectors)" as a write-phase class, with the required defense being *"Validate provenance before write; retain explicit source metadata in long-term memory."* It reports "write-gate validation" as "a shared blind spot across every system examined."

Cortex is one of the systems with that blind spot, and the path is complete end to end.

## Verified path (2026-08-06, on `main` at 65395fb)

**1. Untrusted content selects itself for persistence.** `mcp_server/hooks/post_tool_capture.py:55-58`:
```python
_CONDITIONAL_TOOLS = {
    "WebFetch",
    "WebSearch",
}
```
and `_should_capture()` at lines 101-107:
```python
if tool_name in _CONDITIONAL_TOOLS:
    output_lower = output.lower()
    for kw in _HIGH_VALUE_PATTERNS:
        if kw in output_lower:
            return True, f"keyword:{kw}"
```
The decision to persist is taken **on the fetched content itself**. A page that contains a `_HIGH_VALUE_PATTERNS` keyword is written to the user's long-term store. The attacker controls the predicate.

**2. Novelty gating can be bypassed by shaping the content.** `core/write_gate.py:112` `determine_bypass()` supports `bypass_error` / `bypass_decision` / `bypass_important_tag`, and `docs/mcp-tools.md` states "Decision/error content bypasses the gate." `core/content_cues.py` detects those cues from **content**, multilingually. So text shaped as a decision or an error skips the REJECT verdict that would otherwise suppress it.

**3. It is then replayed verbatim into future sessions.** `hooks/session_start.py` injects anchored + hot memories directly into the context block, with no applicability critique (#363).

**4. And consolidation can raise its confidence.** The same survey's store-phase class is "toxins promoted to higher confidence during summarization/reflection" — Cortex's `core/dual_store_cls.py` episodic→semantic abstraction and `core/compression.py` do exactly this promotion (see #367).

Net: a single fetched page can install a durable, self-reinforcing false "decision" in the user's memory that is replayed into unrelated future sessions. `Bash` is also unconditionally captured (`return True, "bash_output"`), so command output from any source is on the same path.

Note this is not hypothetical for this repo's own workflow — the 2026-08-06 literature sweep that produced this issue ran several `WebFetch` calls against arXiv, each eligible for capture under exactly this rule.

## Why the existing gate does not cover it

`core/abstention_gate.py` is a relevance filter (BEAM-trained; "filters retrieval results that don't actually answer the query") and runs at **read** time. Nothing at write time asks *where did this text come from and is that source trusted for this claim*. §13 D2 requires untrusted data — explicitly including LLM- and network-derived content — to be treated as untrusted wherever the diff consumes it; auto-capture consumes it and promotes it to durable state.

## Ask

1. Stop letting content decide its own persistence: for `_CONDITIONAL_TOOLS`, the capture predicate must not be a keyword match on attacker-controlled output.
2. Record provenance as first-class, queryable metadata on every auto-captured memory (source tool, URL/command, trust class) — not prose inside the content body.
3. Make the write-gate bypass unavailable to untrusted-provenance content: a "decision" cue in fetched web text must not buy the same bypass as a user's deliberate `remember`.
4. Make provenance visible at injection so #363's critique step can act on it, and `/why` can attribute an answer to a network-sourced memory.

## Acceptance criteria

- A test fetches a hostile fixture page containing a `_HIGH_VALUE_PATTERNS` keyword **and** a decision cue, and asserts it is NOT persisted as a trusted memory — with the refusal signal asserted, not just an absent row (§13 A3/F1).
- Provenance/trust class is asserted present and correct on every auto-capture path (`_HIGH_VALUE_TOOLS`, `_LIGHT_VALUE_TOOLS`, `_CONDITIONAL_TOOLS`, `Bash`), and queryable from the store on both SQLite and PostgreSQL (§12.3).
- A deliberate user `remember` retains today's bypass behaviour — regression test, so the fix does not silently make the product worse at its job.
- Zero surviving non-equivalent mutants on changed files (§12).

Relates to #363 (verbatim injection), #367 (compression lineage), #368 (trust-aware retrieval), #366 (deletion completeness), #360 (provenance coverage).

cdeust:
Closes #365.

Fetched web content could install itself in durable, cross-session memory by shaping itself. Two decisions read attacker-supplied text:

1. `hooks/post_tool_capture._should_capture` keyed WebFetch/WebSearch capture on `kw in output.lower()` over `HIGH_VALUE_PATTERNS` — **the fetched page decided whether it was persisted**.
2. `core/write_gate.determine_bypass` grants `bypass_error` / `bypass_decision` from the content itself, so text merely *shaped* like an error or a decision skipped the novelty REJECT.

`hooks/session_start` then replays stored memories verbatim into later sessions, so both are writes to durable state made from untrusted input.

## Why a third concept, not a reuse

`core/capture_origin` resolves origin from the **producing tool name** — known out-of-band at the hook boundary, unforgeable by the payload. The two neighbours could not carry this:

- `core/provenance` grades **reference verifiability** by reading the content's references.
- `core/source_monitoring` attributes **epistemic origin** by regex over the content's wording.

Both are content-derived, so the attacker supplies the classifier's input. **A hostile page dense with file paths and URLs grades `verified` and classifies `perceived` — the most credible value in each vocabulary.** A test asserts the new classifier ignores content entirely.

## Deliberate asymmetries

- `force` and a `deliberate` write class are out-of-band human signals and stay valid at **any** origin. Only the two content-read bypasses are refused, and only for network origin.
- The `important`/`critical` tag bypass also stays, because `_build_tags` never derives those from output — verified, not assumed.
- Unrecognised tools classify `unknown`, never `local_action`, so a newly added tool is visibly unclassified rather than silently trusted.
- `ORIGIN_UNKNOWN` is permissive so no existing caller changes behaviour; every untrusted channel reaches the gate through the classification table, never that default.

## Persistence

New `memories.capture_origin` on both backends with a one-shot migration. **The upgrade path is verified old-code-to-new-code against live PostgreSQL, not reasoned about:** a DB seeded by pre-#365 code (column absent, `ddl_hash cfaedf74`) then initialised by this code gains the column on the TABLE *and* on the `current_memories` VIEW, keeps its pre-existing row, and backfills it to `unknown`.

That view is `SELECT * FROM memories`, whose column list PostgreSQL freezes at creation, so an upgraded DB only sees the column because `get_all_ddl` orders `MIGRATIONS_DDL` before the view. **That ordering is invisible from the fresh-install path** and now has a test.

## Completion Ledger

Test names enumerated from `git diff origin/main...HEAD` — not recalled.

| Diff path | Asserting test |
|---|---|
| network tools → `network` | `test_network_tools_classify_as_network` |
| local tools → `local_action` | `test_local_tools_classify_as_local_action` |
| unknown/empty → `unknown`, never a guess | `test_unrecognised_or_empty_is_unknown_never_a_guess` |
| result always in the vocabulary | `test_result_is_always_in_the_vocabulary` |
| case/whitespace normalised | `test_whitespace_and_case_are_normalised` |
| **classification ignores content (the security invariant)** | `test_classification_ignores_content_entirely` |
| network refused a content bypass | `test_network_origin_may_not_claim_a_content_bypass` |
| non-network permitted | `test_non_network_origins_may` |
| permissive default is a decision on record | `test_unrecognised_origin_is_permissive_by_design` |
| error-shaped content: bypasses locally, refused on network | `test_error_shaped_content_bypasses_at_local_origin`, `test_error_shaped_content_is_refused_at_network_origin` |
| decision-shaped content: same pair | `test_decision_shaped_content_bypasses_at_local_origin`, `test_decision_shaped_content_is_refused_at_network_origin` |
| `force` still wins at network origin | `test_force_still_wins_at_network_origin` |
| `deliberate` still wins | `test_deliberate_write_class_still_wins_at_network_origin` |
| `important` tag still wins | `test_important_tag_still_wins_at_network_origin` |
| pre-change behaviour preserved for existing callers | `test_default_origin_preserves_pre_change_behaviour` |
| capture no longer depends on the payload | `test_same_verdict_regardless_of_keywords`, `test_capture_decision_does_not_depend_on_the_payload` |
| reason string does not echo attacker text | `test_reason_names_the_tool_not_a_keyword` |
| length floor retained | `test_length_floor_still_applies` |
| hook hands the producing tool to `remember` | `test_the_hook_reports_the_producing_tool_to_remember` |
| **hostile payload cannot claim a bypass** (with a precondition that the fixture really trips the detectors) | `test_hostile_payload_cannot_claim_a_content_bypass` |
| hostile payload stored *labelled* network | `test_hostile_payload_is_stored_labelled_network` |
| origin persisted per class | `test_origin_tool_is_persisted_as_its_origin_class` |
| absent tool → `unknown` | `test_absent_origin_tool_persists_unknown` |
| unrecognised tool → `unknown` | `test_unrecognised_tool_persists_unknown_not_a_guess` |
| both base schemas declare the column | `test_pg_base_schema_declares_the_column`, `test_sqlite_base_schema_declares_the_column` |
| migration exists for existing DBs on both backends | `test_pg_migration_adds_the_column_for_existing_databases`, `test_sqlite_migration_list_includes_the_column` |
| **migrations precede the view recreation** | `test_migrations_run_before_the_view_is_recreated` |

### §13.1 checklist

| Item | Evidence |
|---|---|
| A1 | Full suite **7,121 passed, 81 skipped, 123 subtests, 0 failures** (222s, post-rebase) |
| A2 | empty/whitespace/unknown tool, too-short output, hostile payload, shared vocabulary bounds |
| A3/F1 | refusal returns `(False, None)`; capture refusal reason names the tool, not the payload |
| D2 untrusted data | the entire point: network content treated as untrusted at the write boundary |
| E1 | `origin_tool` is a new optional input; no field removed |
| **E3 persisted-data compat** | one-shot migration, no shim; old-data-in/new-code verified on live PG |
| **§12.3 cross-backend** | persistence tests pass on **SQLite (10/10) and PostgreSQL (10/10)** |
| G4 | `test_absent_origin_tool_persists_unknown` asserts absence of a guess |
| G5 | `ruff check` + `ruff format --check` clean |
| H1 | `capture_origin.py` 139 lines, stdlib-only, no `os`/`pathlib`/infrastructure imports |
| H4 | `docs/mcp-tools.md` write path + CHANGELOG |
| **§12 mutation** | 1 survivor, **documented-equivalent** (§12.4): `(tool_name or "")` → `or "XXXX"` is unobservable because an empty and an unrecognised name both return `ORIGIN_UNKNOWN`. Rationale at the use site, refutable |

### Scoped deliberately out

`capture_origin` is queryable by SQL but **not** added to recall results. `_WRRF_CONTRACT_FIELDS` pins the injected-candidate key set to the exact `RETURNS TABLE` column set of the `recall_memories()` PL/pgSQL function; adding the field to the spreading-activation path alone created that divergence and its own contract test caught it. Carrying it onto recall results means changing that stored function's signature and belongs with **#363**, the consumer that needs it, so change and consumer are tested together.

### Also fixed en route

- Removed an unreachable branch: `if tool_name == "Bash"` sat inside the `_CONDITIONAL_TOOLS` block while `Bash` was in `_HIGH_VALUE_TOOLS` — unreachable **in the commit that introduced both** (`5eba011`), so not a leftover and not an unplugged wire. Its intent is already implemented by the high-value branch, which is why deleting it loses no behaviour.
- `origin_tool` exposed on both registered MCP wrappers (the #98 parity gate required it).
- `I2`'s `ALLOWED_WRITERS` pins `heat_base` sites by line number; six entries shifted +4 and were confirmed byte-identical at their new lines before the registry was updated.

### Two false verifications caught

- `psql -c` returns 0 on SQL error without `ON_ERROR_STOP`, so a `DROP COLUMN` that failed on a view dependency reported success — the migration test it produced proved nothing.
- Schema init is guarded by `schema_meta.ddl_hash`, so mutating the DB out-of-band and re-running init skips all DDL. Only a genuine old-code → new-code run exercises a migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,25 @@ adheres to [Semantic Versioning](https://semver.org/).
   added tool is visibly unclassified. `origin_tool` is a declared input on
   `remember` and exposed on both registered MCP wrappers.
 
+  The origin is persisted to a new `memories.capture_origin` column (both
+  backends, with a one-shot migration) so the value that governed the gate is
+  queryable afterwards — an in-flight-only check cannot be audited, and the
+  injection-time critique (#363) and `/why` both need to read it. The upgrade
+  path was verified old-code-to-new-code against PostgreSQL: table and the
+  `current_memories` view both gain the column, pre-existing rows survive and
+  backfill to `unknown`. That view is `SELECT * FROM memories`, whose column
+  list PostgreSQL freezes at creation, so `get_all_ddl` ordering
+  (migrations before the view) is load-bearing and now has a test.
+
+  The value is queryable by SQL on both backends but is deliberately NOT added
+  to recall results yet: `_WRRF_CONTRACT_FIELDS` pins the injected-candidate key
+  set to the exact `RETURNS TABLE` column set of the `recall_memories()`
+  PL/pgSQL function, so surfacing it on the spreading-activation path alone
+  would create the divergence that contract exists to prevent. Carrying it onto
+  recall results means changing that stored function's signature, which belongs
+  with #363 — the consumer that needs it — so the change and its consumer are
+  tested together.
+
   Note the two neighbouring modules deliberately not reused: `core/provenance`
   grades reference verifiability and `core/source_monitoring` attributes
   epistemic origin, both by reading the content. A hostile page dense with file
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,34 @@ adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Security
+
+- **Capture and write-gate bypass are decided by the channel, not the content
+  (issue #365).** Fetched web content could install itself in durable,
+  cross-session memory by shaping itself. Two decisions read attacker-supplied
+  text: `hooks/post_tool_capture` keyed WebFetch/WebSearch capture on a keyword
+  match over the fetched output, so a page decided whether it was persisted; and
+  `core/write_gate.determine_bypass` grants `bypass_error`/`bypass_decision`
+  from the content itself, so text merely shaped like an error or a decision
+  skipped the novelty REJECT. `hooks/session_start` then replays stored memories
+  verbatim into later sessions. Capture for those tools is now a fixed length
+  floor identical for any payload, and the new `core/capture_origin` resolves
+  the origin from the producing tool name — known out-of-band, unforgeable by
+  the payload — so network-origin content is refused the two content-derived
+  bypasses. `force` and a `deliberate` write class are out-of-band human
+  signals and remain valid at any origin; the `important`/`critical` tag bypass
+  also remains, because `_build_tags` never derives those from output.
+  Unrecognised tools classify as `unknown` rather than trusted, so a newly
+  added tool is visibly unclassified. `origin_tool` is a declared input on
+  `remember` and exposed on both registered MCP wrappers.
+
+  Note the two neighbouring modules deliberately not reused: `core/provenance`
+  grades reference verifiability and `core/source_monitoring` attributes
+  epistemic origin, both by reading the content. A hostile page dense with file
+  paths and URLs grades `verified` and classifies `perceived` — the most
+  credible value in each — so neither can carry a security property.
+
+
 ### Fixed
 
 - **`forget` now deletes across every substrate that holds the content
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -116,7 +116,7 @@ table above — the catalogue and the command ship in the same commit.
 
 ### Memory Write Path
 
-1. **Gate**: 4-signal novelty filter (embedding distance, entity overlap, temporal proximity, structural similarity). Decision/error content bypasses the gate — detection is language-aware (see `docs/data-flow.md` § Write Gate Bypass); `force=true` or an `important`/`critical` tag always bypasses, in any language
+1. **Gate**: 4-signal novelty filter (embedding distance, entity overlap, temporal proximity, structural similarity). Content from a NETWORK origin (`origin_tool` = WebFetch/WebSearch — issue #365) is refused the content-derived bypasses, so fetched text cannot skip the gate by looking like a decision or an error; `force` and a `deliberate` write class still bypass at any origin. Decision/error content bypasses the gate — detection is language-aware (see `docs/data-flow.md` § Write Gate Bypass); `force=true` or an `important`/`critical` tag always bypasses, in any language
 2. **Curate**: Active curation — merge with similar, link to related, or create new
 3. **Store**: PostgreSQL + pgvector with auto tsvector indexing → entity extraction → knowledge graph
 
--- a/mcp_server/core/capture_origin.py
+++ b/mcp_server/core/capture_origin.py
@@ -108,9 +108,15 @@ def classify_capture_origin(tool_name: str) -> str:
 
     Never inspects content: the argument is a tool name by construction.
     """
+    # `or ""` is None-safety for .strip(); the empty string then simply matches
+    # neither set and falls through to ORIGIN_UNKNOWN, so no separate
+    # empty-input guard is needed. Documented equivalent mutant (§12.4): the
+    # scoped mutation run replaces this literal with a non-empty one
+    # ("XXXX"), which is unobservable — an unrecognised name and an empty name
+    # both return ORIGIN_UNKNOWN, by construction of the two membership tests
+    # below. No test can distinguish them, so this is equivalent rather than a
+    # coverage gap.
     key = (tool_name or "").strip().lower()
-    if not key:
-        return ORIGIN_UNKNOWN
     if key in _NETWORK_TOOLS:
         return ORIGIN_NETWORK
     if key in _LOCAL_ACTION_TOOLS:
--- a/mcp_server/core/capture_origin.py
+++ b/mcp_server/core/capture_origin.py
@@ -1,0 +1,139 @@
+"""Capture origin — which CHANNEL produced a memory's content (issue #365).
+
+Pure business logic, zero I/O, stdlib only.
+
+This module answers one question: did this content come from the user, from
+this machine, or from off-machine? It answers it from the **tool that produced
+the content**, which is known out-of-band at capture time and cannot be
+influenced by the content itself.
+
+That last property is the whole point, and it is why this is a third concept
+rather than a reuse of either neighbour:
+
+  - ``core/provenance.py`` grades **reference verifiability** — do this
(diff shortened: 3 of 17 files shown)


cdeust:
Reviewed. The core argument is right and well made: a content-derived classifier cannot carry a security property because the attacker supplies its input, so origin has to come from the channel. `provenance.py` grading a hostile page as `verified` and `source_monitoring.py` classifying it `perceived` is the correct justification for a third concept rather than reuse. Cross-backend parity is handled (both `pg_schema` and `sqlite_schema` in the same PR), which is the §12.3 trap this repo has fallen into before.

Coverage is thorough on the intended behaviour: network refused, `force` still wins, `deliberate` still wins, error- and decision-shaped content both refused at network origin.

**One design point I would like a decision on before this is load-bearing.**

The bypass rule is a *denylist*:

```python
_ORIGINS_REFUSED_CONTENT_BYPASS: frozenset[str] = frozenset({ORIGIN_NETWORK})
```

so `UNKNOWN` may bypass, and `test_non_network_origins_may` pins that as intended. Classification itself is fail-safe — an unrecognised tool lands in `UNKNOWN` rather than being promoted to `LOCAL_ACTION`, which the comment explains well. But the *security decision* built on top of it is fail-open: anything not literally named `webfetch` or `websearch` is permitted.

The concrete trigger is not hypothetical. `_NETWORK_TOOLS` hardcodes two lower-cased names against a host whose tool surface changes. If Claude Code adds a third off-machine tool, or renames one (`WebFetch` → `web_fetch` would still match; `WebRead` would not), that content classifies `UNKNOWN` and silently regains the content bypass. The control stops applying with no failing test and no signal — which is the same shape as the issue #365 this PR closes.

Inverting to an allowlist makes it fail closed:

```python
_ORIGINS_ALLOWED_CONTENT_BYPASS = frozenset({ORIGIN_DELIBERATE, ORIGIN_LOCAL_ACTION})
```

The cost is that callers currently relying on the `UNKNOWN` default lose the bypass, so each would need to pass its origin explicitly. `remember` is `DELIBERATE` by definition, so that call site is a one-line change and arguably more honest than inheriting a default.

If you would rather keep the denylist for compatibility, the cheaper mitigation is a drift guard: a test that pins the host's known tool names and fails when the set it was written against changes, so a new web tool is a red test rather than a silent regression.

Not blocking — the PR is a clear improvement over the status quo either way, and the choice between "fail closed now, touch every caller" and "denylist plus a drift guard" is yours. Happy to implement whichever.

Issue #147, opened 2026-07-15
remember: write_class='deliberate' contract violated (below_threshold rejection) and force=true intermittently raises FileNotFoundError


cdeust:
## Observed (2026-07-15, MCP plugin session, PG backend healthy — recall and plain writes working throughout)

Two related write-path defects observed in one session, likely sharing a root cause:

### 1. `write_class: "deliberate"` rejected by the novelty gate — documented contract violated

The tool description states: *deliberate — « NEVER rejected by the gate for low novelty »* (near-duplicates should be merged/linked/superseded by curation, not dropped). Yet:

- `remember(write_class="deliberate")` → `{"stored":false,"action":"rejected","reason":"below_threshold (novelty=0.174, threshold=0.4)"}`
- Same call, different content → `{"stored":false,"action":"rejected","reason":"below_threshold (novelty=0.392, threshold=0.4)"}`
- Also observed: omitting write_class (source-based fallback should classify as deliberate) → same below_threshold rejection.

### 2. `force=true` intermittently raises `FileNotFoundError: [Errno 2] No such file or directory`

- ~12:30: `remember(..., write_class="deliberate", force=true)` → **stored OK**, `reason: "forced"`, memory_id 4334770.
- ~2h later, same session: two consecutive `remember(..., force=true)` calls → `FileNotFoundError: [Errno 2] No such file or directory` (error text suggests checking DATABASE_URL, but PG was demonstrably healthy — non-force writes and recalls succeeded immediately before/after).
- A subagent in the same period hit the same FileNotFoundError on a plain `remember` and fell back to memory-tool.sh.

### 3. Diagnostic clue — `reason: "bypass_error"` on SUCCESSFUL stores

Earlier in the same session, two successful deliberate writes returned `{"stored":true,"action":"stored","reason":"bypass_error", ...}` (memory_ids 4334564, 4334566). `bypass_error` as a reason on a *successful* store suggests the write-class bypass path **throws internally** and the code falls through: sometimes into a plain store (symptom 3), sometimes into the flat gate which then rejects a deliberate write (symptom 1), and on the force path the exception escapes as FileNotFoundError (symptom 2). One failing internal path — probably a missing file/dir dependency (artifact store? gist spool? classifier asset) — with three inconsistent surfacings.

## Expected

- `deliberate` writes are never novelty-rejected (per tool contract) — duplicates go through curation (merge/link/supersede).
- `force=true` always inserts or fails with an actionable error naming the actual missing path.
[...]

PR #380, opened 2026-08-07
fix(capture): make the write-gate bypass rule an allowlist (#365 follow-up)

cdeust:
Follow-up to #375, implementing the review finding. Refs #365.

## What was wrong

The origin control was a **denylist**:

```python
_ORIGINS_REFUSED_CONTENT_BYPASS = frozenset({ORIGIN_NETWORK})
```

Classification already failed safe — an unrecognised tool lands in `UNKNOWN` rather than being promoted to `LOCAL_ACTION`. But the security decision built on top of it failed **open**: anything not literally `webfetch` or `websearch` could claim a content-derived bypass.

The trigger is not hypothetical. `_NETWORK_TOOLS` hardcodes two names against a host whose tool surface changes. Add a third off-machine tool, or rename one, and its content classifies `UNKNOWN` and silently regains the bypass — the control stopping with no failing test and no signal. That is the same shape as the issue #375 closes.

Now an allowlist of `{DELIBERATE, LOCAL_ACTION}`. An unclassified origin is refused, so the cost of a missing classification is a rejected write instead of a trusted one.

## Two things the change forced out

**`ORIGIN_DELIBERATE` was dead.** Defined, documented in the vocabulary, produced by nothing since it was written. Under an allowlist a direct `remember` would resolve `UNKNOWN` and lose the content bypass it has always had, so the value now has a producer: a `remember` carrying no producing tool is the user asking directly.

The condition is the **absence of a tool name**, not an `UNKNOWN` classification. Those differ exactly where it counts — a named-but-unrecognised tool also classifies `UNKNOWN`, and promoting *that* to `DELIBERATE` would reinstate the fail-open this PR removes. My first draft got this wrong and `test_capture_origin_persistence.py::test_unrecognised_tool_persists_unknown_not_a_guess` caught it.

**The issue #147 ordering broke.** A deliberate write's reason degraded from `bypass_error` to the generic `bypass_write_class_deliberate`, masking the diagnostic that test exists to protect. Since such a write bypasses either way, refusing it the specific label changed no outcome and only destroyed information. The origin rule now governs whether content may **buy** a bypass, not how an already-granted one is **labelled**.

## The honest cost

The original design note said `UNKNOWN` was chosen so *"adding this parameter changes no existing caller's behaviour"*. Under an allowlist that is no longer true, and **seven tests in `test_write_gate.py` were silently relying on it**. They now state which channel produced their content — which is what they were asserting all along: *error-shaped content from a local tool* bypasses, not *error-shaped content from anywhere*.

That is the trade in one line: the permissive default was doing real work, and now every caller has to mean it.

## Completion Ledger

| § | Item | Evidence |
|---|---|---|
| A1 | Happy paths | `pytest -k "gate or capture or remember or origin or write_class"` → **453 passed**, 13 subtests |
| A2 | Edge cases | Absent tool, named-but-unrecognised tool, empty string, each origin value, deliberate write class at network origin, important/critical tag at network origin |
| A3 | Failure paths | Refusal asserted directly: `UNKNOWN` refused, unrecognised origin refused, network refused; each with its own test |
| A4 | Trust boundary | The whole change. Origin comes from the channel, never from content; the promotion to `DELIBERATE` keys on tool **absence** and is additionally gated on a write class the auto-capture hook pins out-of-band |
| A5 | Invariants | Stated at the use site: an unclassified origin is refused, and the cost of a missing classification is a rejected write |
| A6 | Idempotency | Pure functions over inputs |
| B1–B3 | Concurrency | N/A |
| C1–C3 | Resources / perf | N/A — set membership |
| D1 | Injection class | N/A |
| D2 | **Untrusted data** | The point of the PR. A future or renamed off-machine tool can no longer buy a content-derived bypass by being unclassified |
| D3 | Secrets | None touched |
| E1 | API compatibility | `may_bypass_write_gate_on_content` keeps its signature; the semantics invert, which is the intended behaviour change |
| E2 | Downstream consumers | `write_gate.determine_bypass` and `handlers/remember`; both updated and covered. Seven test call sites made explicit |
| E3 | Persisted data | `capture_origin` column now records `deliberate` for direct writes where it recorded `unknown`; more accurate, no migration needed |
| E4 | Cross-platform | Pure Python |
| F1 | Signals | The specific bypass reason is preserved for deliberate writes rather than masked |
| F2 | Degraded modes | Refusal is explicit, not silent: `determine_bypass` returns `(False, None)` and the gate reports a rejection |
| G1 | Path→test ledger | Allowlist membership, `UNKNOWN` refusal, unrecognised-origin refusal, absent-tool promotion, unrecognised-tool non-promotion, deliberate-keeps-specific-reason |
| G2 | Regression test | `test_unrecognised_tool_persists_unknown_not_a_guess` is the regression test for the flaw in my own first draft |
| G3 | Determinism | Slice re-run three times; 453 each time |
| G4 | Negative assertions | Named-but-unrecognised is **not** promoted; `UNKNOWN` may **not** bypass; a caller naming no origin gets `(False, None)` |
| G5 | Full gate | `ruff check .` clean · `ruff format --check .` clean (1283 files) |
| H1–H3 | Standards, readability | Each decision states the failure it prevents at the use site |
| H4 | CHANGELOG | Deferred to review: #375's entry describes the denylist. Say the word and I amend it rather than adding a second entry |
| H5 | Commit hygiene | One commit |
| H6 | CI green | Required before merge |
| H7 | Boy-scout (§14) | Found and fixed a dead vocabulary value (`ORIGIN_DELIBERATE`) that predates this PR |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,20 @@ adheres to [Semantic Versioning](https://semver.org/).
   verbatim into later sessions. Capture for those tools is now a fixed length
   floor identical for any payload, and the new `core/capture_origin` resolves
   the origin from the producing tool name — known out-of-band, unforgeable by
-  the payload — so network-origin content is refused the two content-derived
-  bypasses. `force` and a `deliberate` write class are out-of-band human
-  signals and remain valid at any origin; the `important`/`critical` tag bypass
-  also remains, because `_build_tags` never derives those from output.
-  Unrecognised tools classify as `unknown` rather than trusted, so a newly
-  added tool is visibly unclassified. `origin_tool` is a declared input on
-  `remember` and exposed on both registered MCP wrappers.
+  the payload. The two content-derived bypasses are then granted by an
+  **allowlist**: only a `deliberate` or `local_action` origin may claim them.
+  Network content is refused, and so is anything unclassified — a denylist of
+  just the two known network tools would have failed open the moment the host
+  added or renamed one, silently restoring the bypass with no failing test.
+  `force` and a `deliberate` write class are out-of-band human signals and
+  remain valid at any origin; the `important`/`critical` tag bypass also
+  remains, because `_build_tags` never derives those from output. Unrecognised
+  tools classify as `unknown`, which is refused rather than trusted. A
+  `remember` carrying no producing tool *at all* resolves to `deliberate` — the
+  user asked directly — which is deliberately distinct from a tool that was
+  named but is not recognised, since promoting the latter would reinstate the
+  fail-open. `origin_tool` is a declared input on `remember` and exposed on
+  both registered MCP wrappers.
 
   The origin is persisted to a new `memories.capture_origin` column (both
   backends, with a one-shot migration) so the value that governed the gate is
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -116,7 +116,7 @@ table above — the catalogue and the command ship in the same commit.
 
 ### Memory Write Path
 
-1. **Gate**: 4-signal novelty filter (embedding distance, entity overlap, temporal proximity, structural similarity). Content from a NETWORK origin (`origin_tool` = WebFetch/WebSearch — issue #365) is refused the content-derived bypasses, so fetched text cannot skip the gate by looking like a decision or an error; `force` and a `deliberate` write class still bypass at any origin. Decision/error content bypasses the gate — detection is language-aware (see `docs/data-flow.md` § Write Gate Bypass); `force=true` or an `important`/`critical` tag always bypasses, in any language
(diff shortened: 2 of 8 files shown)


cdeust:
Ledger row H4 updated: the CHANGELOG is no longer deferred.

#375's entry described the denylist it shipped — "network-origin content is refused the two content-derived bypasses" and "unrecognised tools classify as `unknown` rather than trusted, so a newly added tool is visibly unclassified". Both were wrong in the same direction once this landed, since `unknown` is now refused rather than merely visible, so the release notes would have contradicted the code.

Amended in place rather than appended to, so there is one account of the behaviour instead of two that disagree. It now states the allowlist, why the denylist was abandoned (it fails open the moment the host adds or renames a network tool), and the distinction between a tool that was *named but unrecognised* and *no tool named at all*.

`docs/mcp-tools.md`'s write-path description carried the same denylist framing and is reconciled too.

Issue #336, opened 2026-08-02
ci: share the test job between ci.yml and release.yml via workflow_call


cdeust:
## The failure class

`ci.yml` and `release.yml` both run the full pytest suite against the same Postgres service, but their trigger sets do not overlap:

```yaml
# ci.yml:4-8
on:
  push:
    branches: [main]
  pull_request:
```

```yaml
# release.yml:48-51
on:
  push:
    tags: [...]
```

**A tag push never reaches `ci.yml`.** So every hardening pass CI receives applies to one copy only, and the two jobs drift apart with nothing to report it. Drift is invisible until a release fails — which is exactly how it surfaced.

## What it cost, measured

Run [30741657854](https://github.com/cdeust/Cortex/actions/runs/30741657854) (tag `v4.17.0`): the `test` job hung on FlashRank's untimeouted `requests.get(..., stream=True)` until pytest-timeout killed the suite, blocking all five downstream publish jobs. **v4.17.0 is tagged and published nothing** — no GitHub release, no PyPI upload, no `.mcpb` bundle — on a tree that had just passed 20 green checks on #334.

Every network-hardening step CI absorbed since 2026-07-27 had skipped `release.yml`. #335 realigned the two by hand and shipped the result as v4.17.1.

## Why the hand-alignment is not the fix

#335 restores parity at one point in time. It does nothing to prevent the next divergence: the two jobs are still two files, and the next person to harden CI still has no signal that a second copy exists. The comment block added at `release.yml:120-129` is documentation, not a mechanism — it can be read past.

## Proposed fix

Extract the shared job into a reusable workflow (`.github/workflows/test-suite.yml`, `on: workflow_call`) that both callers invoke, so hardening lands in one place by construction.

The one asymmetry that must survive the extraction: `requirements/release.txt` deliberately omits tree-sitter, tree-sitter-language-pack, igraph, leidenalg and texttable, so the release gate skips the AST tests and needs none of ci.yml's three tree-sitter steps — porting them blindly fails on `ImportError`. The reusable workflow therefore needs an input (requirements file / extras set) rather than a single hardcoded install, and the narrower release surface should stay a **stated, visible** property rather than an accident of which file you happened to edit.

## Acceptance

- one definition of the test job, called by both `ci.yml` and `release.yml`;
- the release gate's narrower dependency surface expressed as an explicit input, not a duplicated step list;
- a green tag-triggered release run proving the extraction did not break the publish path.

Filed as the follow-up recorded in #335.

Issue #179, opened 2026-07-24
bug: marketplace.json pins withhold downstream releases — zetetic-team-subagents stuck 6 releases back, cortex-viz 3, with no signal


cdeust:
## The defect

`.claude-plugin/marketplace.json` in this repo defines the **`cortex-plugins` marketplace**, which is what `~/.claude/settings.json` actually enables for `cortex@`, `cortex-viz@` and `zetetic-team-subagents@`. It pins each downstream plugin by version. Those pins are stale on `origin/main` today:

| Plugin | Pinned here | Actually released | Lag |
|---|---|---|---|
| `zetetic-team-subagents` | **v2.29.0** | v2.34.0 | 6 releases |
| `cortex-viz` | **v2.6.3** | v2.7.1 | 3 releases |

Nothing updates these pins when the downstream repo releases. Verified 2026-07-24 by fast-forwarding a marketplace clone to current `origin/main` — the pins do not move:

```
$ git -C ~/.claude/plugins/marketplaces/cortex-plugins pull --ff-only origin main
$ python3 -c "... marketplace.json ..."
hypermnesia-mcp=v4.16.0, cortex=v4.15.0, cortex-viz=v2.6.3, zetetic-team-subagents=v2.29.0
```

## Why this is a Cortex issue and not a downstream one

**A downstream release does not reach users until this file changes.** `cdeust/zetetic-team-subagents` cut v2.30.0 through v2.34.0 — the §15 No-Deviation Rule, the redaction quality gates, the Fable multi-model loops — and every one of those releases delivered to **zero installs**, because delivery is gated by a version string here that nothing bumps. Same for three cortex-viz releases.

The downstream repos did everything right: tagged, released, changelogged. The delivery path silently dropped it. That makes this repo the owner of the defect, and the failure mode is the worst kind — it produces no error, no warning, and no diff. A user running `zetetic-team-subagents` believes they are on the current rules; they are six releases behind and nothing says so. For a plugin whose entire product is *rule currency*, that means agents enforce a superseded standard **while reporting compliance** (cdeust/zetetic-team-subagents#52).

This is the §13 F2 class exactly: a degraded mode that is not explicit, not named in the output, and silently defaulted to.

## Immediate remediation

Bump both pins to the current releases (`zetetic-team-subagents` v2.34.0, `cortex-viz` v2.7.1) and verify each loads.

## Acceptance criteria — the pins must stop going stale

[...]

PR #395, opened 2026-08-08
docs: record the CI-drives-the-release inversion (ADR-0055)

cdeust:
Phase 5 (last) of #392. **Stacked on #394**, which is itself stacked on #393. Merge order: #393 → #394 → this.

## Why

`CLAUDE.md` § *Releasing* referenced "the release checklist" — a document that **does not exist anywhere in the repo**. It only stated what the checklist *ends* with, and named 2 of the 14 version sites a release actually has to move. It also predates #392: it implied a tag is something a human pushes.

## What changes

**`CLAUDE.md` § Releasing** becomes the real procedure, which is now one PR: bump the 14 sites → merge → `release-gate` tags the SHA that went green. The site list is delegated to `scripts/check_version_surfaces.py` rather than restated, so it cannot go stale in prose the way the old text did.

The marketplace-pin paragraph stays. Issue #179's lesson is unchanged, and `check_marketplace_pins.py` still covers what the version gate does not — pin staleness against the latest tag, and the third-party plugin pins.

**`docs/adr/ADR-0055`** records why the tag became an output rather than an input, and — the part worth having in six months — **three rejected alternatives with the evidence that killed each**:

- **`workflow_call`** (the shape #336 originally proposed, and the obvious one): PyPI Trusted Publishing does not support reusable workflows, *and* PR #387 already demonstrated in this very repository that a job delegating via `uses:` renames its status check to `<caller> / <callee>` — all four legs green, PR BLOCKED, because the contexts branch protection required were reported by nobody.
- **Delete the test job and keep tagging manual**: the status quo minus its only safety net.
- **`repository_dispatch`**: avoids the `GITHUB_TOKEN` restriction without a new credential, but severs the tag from the release — `mcpb-bundle` checks `manifest.json.version` against `GITHUB_REF_NAME`, `github-release` derives its notes from the tag range, and three attach steps use `${{ github.ref_name }}`.

## Verification

```
scripts/check_doc_claims.py        → doc claims OK
scripts/check_version_surfaces.py  → 14/14 agree
scripts/check_ci_gate_complete.py  → CI gate completeness: OK
scripts/check_marketplace_pins.py  → All marketplace pins current.
```

No CHANGELOG entry: CI/tooling-only changes do not get one in this repo (#336, #387, #388 and #391 have none either). There is no ADR index to update — ADRs are referenced from CHANGELOG entries and from the code that implements them, not from a list.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,9 +26,11 @@ name: Release
 # The test guarantee now comes from `ci.yml`'s `release-gate` job: it runs
 # ONLY after `ci-green` succeeds on a push to `main`, and tags the EXACT SHA
 # that just went green with an annotated `git tag` pushed via
-# `RELEASE_TAG_TOKEN` (a `GITHUB_TOKEN`-pushed tag does not start a new
-# workflow run, per GitHub's docs — this workflow would never fire from
-# one). The released tree is therefore bit-identical to the validated one;
+# the `RELEASE_TAG_SSH_KEY` deploy key (a `GITHUB_TOKEN`-pushed tag does not
+# start a new workflow run, per GitHub's docs — this workflow would never
+# fire from one; that a deploy-key push DOES fire it is untested, see
+# docs/adr/ADR-0055-ci-drives-the-release.md).
+# The released tree is therefore bit-identical to the validated one;
 # this workflow's only job is to propagate what `ci.yml` already validated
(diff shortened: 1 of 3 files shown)


cdeust:
Fermée à la demande du mainteneur. Le travail sera refait de manière à ne pas entrer en conflit avec la CI et la protection de branche existantes.

PR #399, opened 2026-08-09
feat(recall): trust/provenance term in WRRF fusion (#368)

### LINKED ISSUE #368: recall fuses relevance and heat only — no trust or provenance dimension in WRRF, so a poisoned memory outranks a good one
The retrieve-phase class in [arXiv 2604.16548](https://arxiv.org/abs/2604.16548) is retrieval corruption: *"RAG poisoning (malicious entries ranked highest by embedding similarity)"* and *"memory entries overriding explicit user instructions."* Required defense: a *"trust-aware or provenance-aware retrieval policy"* with consensus validation. The survey is explicit that the read-side filter Cortex relies on is not enough: **"Retrieval-time filtering alone is insufficient"**, and that existing defenses *"have not been stress-tested by adaptive LLM-driven red-teaming."*

## Current state (verified 2026-08-06)

Per `docs/mcp-tools.md`, the read path fuses **vector + FTS + trigram + heat + recency** via the PL/pgSQL `recall_memories()` WRRF, then reranks with FlashRank (`core/reranker.py`), then filters with `core/memory_rules.py` and `core/abstention_gate.py`.

Every one of those signals is a **relevance or salience** signal. None is a trust signal:

- `core/thermodynamics.py` heat rises with access — so a memory that keeps getting retrieved keeps ranking higher, which rewards a successful poisoning rather than penalising it.
- `core/reranker.py` scores query-passage relevance; a well-crafted malicious passage is *more* relevant, not less.
- `core/abstention_gate.py` asks "does this answer the query" — the poisoned entry does.
- `core/memory_rules.py` hard/soft rules are user-authored and static; they cannot express "distrust anything whose provenance is a fetched web page."

So the ranking function has no term that could ever demote a memory for being untrustworthy, and the heat term actively amplifies whatever gets retrieved.

## Ask

Add a trust/provenance term to retrieval, consuming the attribute #365 introduces at write time.

1. A provenance-derived trust factor in the fusion, not only as a post-filter — the survey's point is that filtering after ranking is insufficient.
2. Break the heat feedback loop for low-trust provenance: retrieval of an untrusted memory should not compound its future ranking.
3. Red-team the ranker: adversarial passages crafted to win on similarity, per the survey's note that defenses are untested against adaptive attacks.

## Acceptance criteria

- A test seeds a high-similarity, low-trust memory alongside a lower-similarity, high-trust one and asserts the ordering the policy intends — on both backends (§12.3).
- The trust term's weight is **sourced** (§8): a benchmark under `benchmarks/` or a dated measurement, not a chosen constant. No invented constants.
- Retrieval quality on the existing suites is measured before/after with paired arms at one commit; a relevance regression is stated and justified or the change does not ship.
- An adversarial-passage fixture set is committed and exercised in CI.
- Zero surviving non-equivalent mutants on changed files (§12).

Relates to #365, #366, #363, #360.

cdeust:
Closes #368 (when the calibration lands — see *Not ready to merge* below).

## What this adds

A trust/provenance term in the WRRF fusion, so a candidate whose origin is
untrusted can be demoted at ranking time. The mechanism is complete and green
on both backends, and **inactive in production**: `UNTRUSTED_ORIGIN_FACTOR`
defaults to `1.0` (identity — it demotes nothing) until the calibration below
picks a value.

Entry measurement: 4/4 adversarial attacks succeeded. Cheap pre-sweep:
W=1.0 → 0/4 defended, 0.95–0.80 → 2/4, W ≤ 0.70 → 4/4.

## Commits

- `8d4e92fe` — the mechanism
- `64a35ca8` — split `pg_recall.py` at its two documented seams (the
  craftsmanship gate blocked on pre-existing debt in a file this touches)
- `5177cf32` — sweep: empty-array expansion under `set -u` (bash 3.2)
- `66d2628f` — **benchmark harness fix**, see below

## The harness fix is the load-bearing part of this push

`reproduce.sh::start_db()` waited on `pg_isready` over the container's Unix
socket. The postgres entrypoint runs initdb against a socket-only temporary
server (`listen_addresses=''`), so the socket reports ready *during*
initialization and the benchmark's own connection from the host is refused.
It killed 3 of 5 calibration cells in ~3s each. Loud failures (rc=1) — no cell
produced wrong numbers, only no numbers.

Probing TCP inside the container was tried first and **rejected by
measurement** (optimistic too, median 2.06s, 5/5 runs). The probe is now the
real thing: a psycopg connection from the host over the published port.

Full measurement, the entrypoint evidence, and the two errors made *while
measuring* (ordering bias; locale-truncated awk) are recorded in
`docs/provenance/pg-readiness-probe-2026-08-09.md`.

This dragged `reproduce.sh` further past the §4 size limits it already
violated, so per the boy-scout rule it was split: container lifecycle →
`benchmarks/lib/bench_container.sh`, MANIFEST heredoc →
`benchmarks/lib/write_manifest.py` (now lint-checked like the Python it always
was). 621 → 440 lines, no function over 50. Verified behaviour-preserving:
identical scores (MRR 0.8439, R@10 0.9746) and an identical MANIFEST on every
non-volatile field.

**This is why CI matters on this PR** — it exercises the release gate harness
itself.

## Not ready to merge

The calibration grid {1.0, 0.8, 0.7, 0.6, 0.5} is **re-running now** (~7h) on
the corrected harness. The first attempt is discarded in full — including the
cell that completed — because a grid cannot mix provenances.

Remaining before this leaves draft:
1. apply the pre-registered rule (largest W defending 4/4 with all four
   `reproduce.sh` floors intact) — unchanged, not reinterpreted
2. fill §Results in `docs/provenance/trust-factor-calibration.md`
3. wire that W with a `# source:` pointing at the pre-registration
4. if no W satisfies both, **no floor gets relaxed** — it comes back for a
   decision

Out of scope, filed rather than fixed: `pg_store.py` is 1384 lines.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -102,6 +102,28 @@ a downloaded checksum. The container image built in CI is a bare-container
 *smoke test* (`ci.yml`, `docker-smoke`) — it is never pushed to a registry,
 so there is no published image to attest.
 
+### Dismissed CodeQL alerts
+
+An alert is dismissed here only when the dismissal is *checkable* — the
+paragraph must state which value CodeQL believed was sensitive and where that
+value is defined, so a reader can refute the dismissal without rerunning the
+scanner. Anything else stays open.
+
+- **#143 — `py/clear-text-storage-sensitive-data`, high severity**
+  (`benchmarks/lib/trust_factor_sweep.py:179`, PR #399). Dismissed as a false
+  positive. The alert names its own two sources: the frozenset at
+  `mcp_server/core/capture_origin.py:131-133` and the `trusted_origins_at_read()`
+  call at `benchmarks/lib/trust_factor_sweep.py:174`. Their value is the triple
+  of public string literals `("deliberate", "legacy", "local_action")`, defined
+  at `capture_origin.py:66-70` — the read-time trust allowlist of issue #368,
+  not a credential. It is written to a benchmark artefact that records which
+  origin policy the sweep ran under, which is provenance, not disclosure. The
+  classification is driven by the identifiers carrying the word *trusted*, and
+  CodeQL itself classifies the alert as `test`. Nothing was renamed to satisfy
+  the scanner: tuning production identifiers to a name heuristic would trade a
+  real signal for silence. Re-open trigger: any change that makes a
+  `trusted_origins*` value non-constant or reader-supplied.
+
 ## Reporting a Vulnerability
 
 If you discover a security issue in this project, **do not** open a public
--- a/benchmarks/lib/adversarial_corpus.py
+++ b/benchmarks/lib/adversarial_corpus.py
@@ -1,0 +1,208 @@
+"""Adversarial retrieval-poisoning corpus (issue #368).
+
+Pure data + construction helpers, zero I/O. Imported by both the test suite
+(`tests_py/`) and the trust-weight ablation, so the passages that prove the
+attack and the passages that measure the defence are the same passages.
+
+Source: arXiv 2604.16548, *retrieve-phase* threat class — "RAG poisoning
+(malicious entries ranked highest by embedding similarity)" and "memory
+entries overriding explicit user instructions". The survey's operative claim
+for this module is that "Retrieval-time filtering alone is insufficient": a
+fixture that only checked a post-ranking filter would not test the property
+under defence.
+
+Design constraint that shapes every passage here: an adversarial passage must
+be **more** retrievable than its legitimate counterpart, never less. A corpus
+whose hostile entries lose on similarity would pass against a system with no
+defence at all, and would therefore measure nothing. Each pair below is
+declared with the adversarial similarity strictly above the legitimate one,
+and `assert_pair_is_adversarial` enforces that at construction time rather
+than trusting the literals to stay ordered as the file is edited.
+
+The payloads are inert by construction — they are strings stored in a test
+database and scored by a ranking function. Nothing here is executed, fetched,
+or interpreted; the hostile property under test is *rank*, not effect.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+# ── Capture origins (mirrors mcp_server/core/capture_origin.py) ────────────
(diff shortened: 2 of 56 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Review performed 2026-08-09 by the zetetic code-reviewer agent, in a fresh context, on `git diff origin/main...feat/issue-368-trust-provenance-wrrf` (7 commits). Posted here because a verdict that lives only in a chat transcript is not a record — and because this PR was previously described as ready on the strength of a green CI alone, which is not a verdict of validity.

## Verdict

**Valid as-is.** Every sourced claim checked — the `# source:` chain from `mcp_server/core/retrieval_dispatch.py:23-41` through `docs/provenance/trust-factor-calibration.md` to the committed artefacts under `benchmarks/results/trust-factor-sweep/` — traced to real data matching what the comment asserts. The specific vice rejected on PR #403 (a `# source:` citing a path that exists only on the developer's machine) does **not** recur here.

The adversarial corpus is a genuine test, not a staged result: Gram-Schmidt-constructed embeddings run through the live ranking path, four distinct attack families, defended counts re-measured and reproduced twice. The calibration document is unusually candid — it states its own limitation rather than smoothing it over, and corrects an earlier drafting error about `--with-consolidation` instead of quietly editing it away.

## The caveat that governs how this should be read

The gated floors prove **non-regression under uniform capture origins**, not the absence of relevance cost under mixed ones. The harnesses never populate `capture_origin`, so W multiplies every row alike and cannot change any ordering — which is why LongMemEval is identical to four decimals. The question this PR settles is narrower than it appears to settle. That is disclosed in the calibration doc, and now has a follow-up: #405.

## Non-blocking findings

- `mcp_server/core/pg_recall.py` came out of its split at 494 lines: under the 500-line global cap, over this repo's stricter 300-line local cap, which the split commit's own message did not cite. Filed as #404.
- The mixed-origin limitation had no follow-up issue. Filed as #405.

Neither blocks the merge.

PR #410, opened 2026-08-09
fix(memory-ingest): stop dropping capture_origin at ingest (issue #368)

cdeust:
## Summary

`mcp_server/core/memory_ingest.py::ingest_memory` accepted a caller-supplied
`capture_origin` field but never forwarded it to `store.insert_memory`,
unlike every other metadata field (`source`, `tags`, `heat`, ...). Every
`ingest_memory` caller therefore silently fell through to `insert_memory`'s
`"unknown"` default regardless of what it intended. **This is a production
code defect, not a harness gap** — a field accepted then silently dropped.

This is why `docs/provenance/trust-factor-calibration.md`'s gated arm could
not discriminate the trust-factor weight W: LME/LoCoMo/BEAM all landed on
`capture_origin='unknown'`, which `core.capture_origin.trust_factor` demotes
*uniformly*, and a uniform multiplier cannot change WRRF order — hence LME
identical to four decimals across W in {1.0, 0.7, 0.6, 0.5} in that document's
existing results table.

## What's fixed

- `memory_ingest.py` — `capture_origin` now passes through like every other
  field. Default unchanged (`"unknown"` for callers that omit it) — no
  behavior change for any existing caller that doesn't set it.
- `benchmarks/lib/capture_origin_mix.py` (new) — the realistic production
  mixture of `capture_origin` values (`local_action` .966 / `network` .025 /
  `deliberate` .009), sourced from a measurement over this machine's own 886
  Claude Code session transcripts (`~/.claude/projects/**/*.jsonl`) — the
  only representative "existing store" available (the local `memory.db`
  predates the `capture_origin` column, no live production DB touched).
  Deterministic sampler, documented derivation and limitation in the module
  docstring.
- `benchmarks/lib/bench_db.py` — wires the mixture into `load_memories()` via
  `_apply_capture_origin_mix()`, so LME/LoCoMo/BEAM now insert a realistic
  blend of trusted/untrusted origins instead of an all-`unknown` corpus. Never
  overwrites an explicit origin already set by a caller (e.g. the adversarial
  corpus).

## What's NOT fixed yet (next, separately)

The production value **W = 0.7 still rests on the invalidated gated arm**
documented in `docs/provenance/trust-factor-calibration.md` — that document
is unchanged by this PR. Re-running the 5-cell W sweep
(`benchmarks/trust_factor_sweep.sh`) under the now-mixed origins is the
direct continuation of this same contract, not a separate task, but it costs
~7h across 5 sequential Docker-isolated `reproduce.sh` cells and that cost is
being reported to the repo owner for a go/no-go before running it — plus a
check for whether the adversarial arm (`benchmarks/lib/trust_factor_sweep.py`,
seconds on in-memory SQLite) is now sufficient on its own, since the origins
it discriminates against are the same ones the gated arm was blind to before
this fix.

## Test plan

- [x] `uv run --no-sync pytest -q -p no:randomly` — 7291 passed, 142 skipped,
      0 failed
- [x] 16 new tests: `tests_py/core/test_memory_ingest_capture_origin.py`,
      `tests_py/benchmarks/test_capture_origin_mix.py`,
      `tests_py/benchmarks/test_bench_db_capture_origin.py`
- [x] `ruff check` / `ruff format --check` clean on touched files
- [ ] CI green (watching)

### DIFF
--- a/benchmarks/lib/bench_db.py
+++ b/benchmarks/lib/bench_db.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 import os
 from typing import Any, Callable
 
+from benchmarks.lib.capture_origin_mix import assign_capture_origins
 from mcp_server.core.memory_ingest import ingest_memories_batch
 from mcp_server.core.pg_recall import (
     assemble_context as pg_assemble_context,
@@ -23,6 +24,23 @@ from mcp_server.core.pg_recall import (
 from mcp_server.core.reranker import ensure_reranker_loaded
 from mcp_server.infrastructure.embedding_engine import EmbeddingEngine
 from mcp_server.infrastructure.pg_store import PgMemoryStore
+
+
+def _apply_capture_origin_mix(memories: list[dict[str, Any]]) -> None:
+    """Stamp every memory lacking `capture_origin` with a value drawn from
+    the measured production mixture, in place.
+
+    Extracted as a free function (rather than inlined in `load_memories`) so
+    it is testable without a live PostgreSQL connection — see
+    `tests_py/benchmarks/test_bench_db_capture_origin.py`.
+
+    A memory that already sets `capture_origin` (e.g. the adversarial-corpus
+    montage, which needs specific per-pair values, not the aggregate mix) is
+    left untouched — `setdefault` only fills what's missing.
+    """
+    origins = assign_capture_origins(len(memories))
+    for mem, origin in zip(memories, origins, strict=True):
+        mem.setdefault("capture_origin", origin)
 
 
 class BenchmarkDB:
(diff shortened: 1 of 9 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Reviewed the diff directly, not the description.

**The production change is one field, and it is the right one.** `mcp_server/core/memory_ingest.py` now threads `capture_origin` through the same dict that already carries `source`, `tags`, `heat`, `is_protected`, `agent_context` and `is_global`. The field was being dropped silently, so every caller — the LME, LoCoMo and BEAM harnesses included — fell through to `insert_memory`'s `"unknown"` default no matter what it set. That is a root-cause fix at the point where the value was lost, not a patch at the reading end.

**Backward compatibility is preserved by construction**: `memory.get("capture_origin", "unknown")` gives an omitting caller exactly the previous behaviour, so nothing downstream can observe a change unless it was already setting the field and being ignored.

**The consequence is stated honestly in the code comment**: this is why the trust-factor calibration's gated arm could not discriminate W. A one-line data-plumbing defect invalidated a measurement — worth recording at the call site, which the comment does, with a pointer to the provenance document rather than a bare assertion.

**The rest of the diff is test and harness surface** (`benchmarks/lib/*`, `tests_py/benchmarks/*`, `benchmarks/reproduce.sh`) — 642 insertions against 1 deletion, i.e. almost entirely new coverage for the behaviour being fixed, including a machine-load manifest writer and a capture-origin mix helper. New tests around a fix are what makes it a regression fix rather than a change.

CI: 22 pass, 2 skipping. No blocking findings.

Issue #282, opened 2026-07-30
mutation: 29 @dataclass classes in mcp_server/ carry methods invisible to mutmut (same defect class as issue #262)


cdeust:
## What

Sweeping the whole repository (not just `scripts/`) for the dataclass-method
mutation-exclusion defect that issues #262 / this branch's PR fixed for
`RepoBadge` (`scripts/generate_repo_badges.py`) and `Ranking`
(`scripts/refresh_mcp_toplist_badge.py`) — mutmut's mutation generator
categorically skips the body of any `@dataclass`-decorated `ClassDef`
(`mutmut/mutation/file_mutation.py:236`, confirmed in the installed
source) — surfaces **29 more decorated classes with non-dunder methods**,
all in `mcp_server/` (the actual runtime package, not the `scripts/` gate
family this PR/#262/#280 have been working through):

```
mcp_server/doctor_mcp.py::McpReport [required_fails, warnings, to_dict]
mcp_server/core/value_learning.py::ValueUpdate [as_dict]
mcp_server/core/sensory_buffer.py::BufferItem [to_dict]
mcp_server/core/goal_maintenance.py::GoalVector [is_active, as_dict]
mcp_server/core/document_model.py::DocumentSection [is_empty]
mcp_server/core/document_model.py::ParsedDocument [is_empty]
mcp_server/core/document_model.py::DocumentProvenance [dedup_tag, source_tag]
mcp_server/core/wiki_coverage.py::DomainCoverage [covered_count, missing_count, coverage_ratio, missing_scopes]
mcp_server/core/wiki_coverage.py::FileCoverage [coverage_ratio]
mcp_server/core/procedural_memory.py::ActionStep [key]
mcp_server/core/procedural_memory.py::ProceduralSkill [is_habitual, length, as_dict]
mcp_server/core/wiki_redirect.py::Redirect [is_id_based]
mcp_server/core/wiki_groomer.py::PageAudit [has_issues]
mcp_server/core/wiki_axis_registry.py::AxisRegistry [values, names, get, has, default_for]
mcp_server/core/ablation.py::AblationConfig [is_enabled, disable, enable, disable_all_except]
mcp_server/core/conflict_monitor.py::ConflictAssessment [as_dict]
mcp_server/core/habituation.py::HabituationOutcome [as_dict]
mcp_server/core/attentional_control.py::AttentionAllocation [as_dict]
mcp_server/core/forward_model.py::ForwardModelState [as_dict]
[...]

Issue #276, opened 2026-07-30
headless-authoring worker family: methods exceed 50-line cap (§4.2), CLAUDE.md size-limit text drifted from CONTRIBUTING.md


cdeust:
## Context

Discovered while fixing #237 (the consolidation-package import cycle). Not
in that issue's blast radius (no test, build, or lint step #237's fix
triggers checks these numbers — confirmed no active pre-commit hook or CI
job enforces file/method size in this repo today), so deferred here per
coding-standards.md §14.3 rather than folded into that PR.

## Finding 1 — method-length debt in the headless-authoring worker family

`mcp_server/handlers/consolidation/{drain_operations,cycle_orchestration}.py`
have several methods well over the §4.2 50-line cap (measured on the
pre-#237 tree, i.e. before that fix's small sentinel-resolution additions):

```
drain_operations.py::drain_one                       ~85 lines
drain_operations.py::drain_all_gaps_on_page           ~130 lines
drain_operations.py::drain_missing_anchors            ~90 lines
cycle_orchestration.py::run_headless_authoring_cycle  ~175 lines (incl. 2 nested coroutines)
```

These were already over the cap before #237 touched these files (that PR
added ~5 lines of sentinel-resolution logic to each, not the underlying
size). Fowler catalog fixes: `Extract Function` on `drain_all_gaps_on_page`'s
per-gap fill loop and `run_headless_authoring_cycle`'s two nested
coroutines (`drain_anchor_bounded`/`drain_page_bounded` could become
module-level functions taking `sem`/`budget` as explicit parameters).

## Finding 2 — CLAUDE.md vs CONTRIBUTING.md disagree on the numbers

- `CLAUDE.md` § Code Style: "300 lines max per file; 40 lines max per
  method. Enforced by the craftsmanship-checker pre-commit hook."
- `CONTRIBUTING.md` § Definition of Done (or similar): "§4.1 File ≤500
  lines, §4.2 function ≤50 lines" — citing coding-standards.md's actual
  §4 numbers.
- No `craftsmanship-checker` pre-commit hook is actually installed
  (`.git/hooks/` has only the standard `*.sample` files), and no CI job
  greps for file/method line counts.

Before splitting the functions in Finding 1, resolve which number is
authoritative (500/50, matching coding-standards.md and CONTRIBUTING.md,
or a locally-tightened 300/40) and fix whichever doc is stale, so the next
split targets the real threshold instead of guessing.

## Acceptance criteria

[...]

Issue #253, opened 2026-07-29
pyright gate resolves a different tree-sitter-language-pack than the documented dev install


cdeust:
## What

The pyright gate resolves a different `tree-sitter-language-pack` than the
documented developer install, so a contributor sees one diagnostic that CI never
reports.

## Measurement (2026-07-29, worktree at f606c37)

| Environment | How it is built | `tree-sitter-language-pack` | `pyright 1.1.410 mcp_server/` |
|---|---|---|---|
| CI (`ci.yml:430-437`) | `python -m venv .venv` + `pip install -e ".[dev,postgresql,sqlite,codebase,otel]" flashrank` | 1.13.5 | `0 errors, 0 warnings, 0 informations` |
| Developer (`CLAUDE.md:21`, `uv sync`) | resolved from `uv.lock` | 1.6.2 | `1 error` |

The diagnostic:

```
mcp_server/core/ast_parser.py:88:23 - error: Argument of type "str" cannot be
assigned to parameter "name" of type "SupportedLanguage" in function "get_parser"
```

Confirmed pre-existing by a paired control: it reproduces on an unmodified
`origin/main` tree in the same venv, and disappears on the same tree in the
CI-equivalent venv.

## Why this matters

`pyproject.toml:77` pins `tree-sitter-language-pack>=0.24.0,<1.14`, a range wide
enough that the two install paths land 7 minor versions apart. The pyright gate
is a zero-diagnostic gate (issue #197); a gate whose result depends on which
installer you used cannot be trusted in either direction — today it costs a
contributor a phantom error, tomorrow it can hide a real one.

Filed per coding-standards §14.3: outside the #239 DTZ family's blast radius —
`ast_parser.py` is untouched by that diff and the divergence is an environment
resolution issue, not a source defect.

## Acceptance criteria

1. The pyright job and the documented developer install resolve the same
   `tree-sitter-language-pack` (either CI installs from the lockfile, or the
   lockfile is refreshed and the pin narrowed).
2. `pyright mcp_server/` reports `0 errors` in BOTH environments, evidenced by
   quoted output from each.
3. `mcp_server/core/ast_parser.py:88` narrows `language` to the
   `SupportedLanguage` literal, or documents why the widening is safe.

Issue #237, opened 2026-07-28
Circular import: four consolidation submodules cannot be imported first (headless_authoring hub)


cdeust:
## Reproduction

Each of these fails in a fresh interpreter (verified on main, 3a1ba1e, 2026-07-28):

```
$ python -c "import mcp_server.handlers.consolidation.candidate_scan"
ImportError: cannot import name '_collect_anchor_candidates' from partially
initialized module ... (most likely due to a circular import)
```

Same failure shape for `claude_cli` (`_build_argv`), `cycle_orchestration`
(`run_headless_authoring_cycle`), and `drain_operations`
(`drain_all_gaps_on_page`).

## Cause

The four submodules each do `from . import headless_authoring as _root` at
module top while `headless_authoring` imports names back from them — the
same load-order-sensitive hub cycle as #233's synaptic pair, one package
over. Importing `headless_authoring` first resolves it, which is why the
suite stays green; any consumer (REPL, script, future module) that imports
a submodule directly fails. `tests_py/handlers/test_s110_sweep_handlers.py`
already has to import `headless_authoring` first with a comment noting the
load-order sensitivity.

Surfaced by the #197 family-4 (PLC0415) sweep's per-module import
verification, which imports every `mcp_server` module in a fresh
interpreter. The sweep touched these files (hoisting unrelated lazy
imports) but deliberately did not restructure the hub cycle
(behavior-preserving mandate; a seam-creating refactor is its own PR per
coding-standards §15.1). The sweep's baseline check proves these four
failures are pre-existing on main, byte-for-byte identical before and
after the hoist.

## Acceptance criteria

1. `python -c "import mcp_server.handlers.consolidation.<m>"` succeeds in
   a fresh interpreter for all four modules (regression test iterating them
   in subprocesses, as #233 criterion 1).
2. The `_root` back-reference is replaced by a leaf module (or explicit
   parameter passing) so no submodule imports the hub at load time.
3. Existing suite passes unchanged.
4. The load-order comment/workaround in
   `tests_py/handlers/test_s110_sweep_handlers.py` becomes unnecessary and
   is removed.

Refs #233 (same defect class), #197 (surfaced by family 4).

PR #412, opened 2026-08-09
fix(tests): drop unsourced 30s subprocess timeout (issue #402)

### LINKED ISSUE #402: Order-dependent failure: test_benchmark_db_resolves_with_psycopg_present fails under random test ordering

cdeust:
## Summary

Issue #402 hypothesized an import-state leak causing `test_benchmark_db_resolves_with_psycopg_present` to fail under randomized full-suite ordering. **That hypothesis is refuted, not fixed around:**

- `subprocess.run([sys.executable, "-c", ...])` starts a **fresh interpreter process**. Nothing in the parent's `sys.modules` can cross that boundary — only `os.environ` and `cwd` are inherited, and I verified the import doesn't depend on `cwd` (works identically from `/tmp`).
- Every `sys.modules`/`os.environ` mutation site across `tests_py/` was audited (agent_briefing, preemptive_context, guard_blocks_populated_db, guard_against_populated_db, pg_schema_provision, pg_throwaway_db, embedding_engine, ast_parser×2, temporal_normalize, otel_exporter, migrate, launcher_deps) — all use `monkeypatch`/`mock.patch.dict` with guaranteed restore, and pairwise testing against the target found zero reproductions.
- 8+ clean full-suite runs (before machine access was withdrawn for an unrelated 7-hour benchmark) did not reproduce the original failure.
- The one confirmed failure traces to a run made on 2026-08-09 while multiple other agent sessions shared the host (load average 14 on 10 cores) — an environmental measurement taken on a saturated machine, not a code-level ordering defect. (Also discovered along the way: `pytest-randomly` is not even a dependency of this repo — collection-order variance comes from OS-level filesystem enumeration, not a seeded plugin, so there was never a seed to bisect against in the first place.)

## What's actually fixed

The test's `timeout=30` on the subprocess call had no `# source:` justification (coding-standards.md §8) — an invented constant whose value silently decided pass/fail under load. Measured baseline for the snippet: **0.48s–0.92s** across 5 runs on a quiet machine (`.venv/bin/python3 -c "..."`, timed 2026-08-10) — already a >30x margin, yet still exceeded under real contention. Enlarging the constant further doesn't fix the underlying defect: any fixed wall-clock bound sized for a quiet machine can be exceeded by an arbitrarily busy one.

Removed the local `timeout=` entirely. A genuine hang is still caught by pytest's own per-test watchdog (`pyproject.toml` `[tool.pytest.ini_options] timeout = 300`, itself sourced to the 2026-05-25 CI stall incident) — reusing that already-sourced, already-relied-upon backstop instead of inventing a second, narrower, unsourced one.

## Test plan

- [x] `tests_py/benchmarks/test_lib_init_no_psycopg.py` — 5 consecutive runs, all green (`4 passed` each, ~2.4-2.5s)
- [x] `ruff check` clean on the changed file
- [x] `ruff format --check` clean on the changed file
- [ ] Full-suite reproduction was explicitly **not** attempted per instruction — the machine was reserved for an unrelated 7-hour benchmark measurement during this investigation

Closes #402 (already closed by the owner as "should never have been opened" — this PR is the promised follow-through, not a reopen).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,28 @@ delegating steps costs nothing.
 `scripts/check_ci_gate_complete.py` (run by `lint`) fails the build
 otherwise — an ungated job could fail without blocking a merge.
 
+### Issue ownership
+
+**Only the repository owner opens issues on this repo.** If you (human or
+agent contributor) find a defect while working — including one outside the
+blast radius of your current change — you do not file a ticket for it.
+
+This is not a license to ignore it. The rule closes the exit, it does not
+open one: a defect you can fix in the material you are already touching
+gets fixed in the same PR (see the boy-scout discipline under Testing/What
+NOT to do below); a defect outside that blast radius gets described plainly
+in the PR description instead of being silently dropped or waved through
+with a "pre-existing" / "unrelated" / "out of scope" label — those labels
+are not a substitute for either fixing it or naming it where a reviewer can
+see it. **A violation you declare in a commit message or a PR description
+is still a violation** — enumerating it does not authorize leaving it in
+place; it is triage information for the owner, who decides whether it
+becomes its own issue, its own PR, or gets folded into the current one.
+
+Concretely: don't invoke "no issues" as a reason to decline fixing or
+reporting something you found. It is a rule about *who opens tickets*, not
+about *what gets addressed*.
+
 ---
 
 ## Adding a biological mechanism
--- a/tests_py/benchmarks/test_lib_init_no_psycopg.py
+++ b/tests_py/benchmarks/test_lib_init_no_psycopg.py
@@ -85,6 +85,27 @@ def test_benchmark_db_resolves_with_psycopg_present():
     postgresql extra") -- there psycopg's absence is the environment, not
     a regression, and `test_benchmark_db_still_reachable_lazily_and_fails_
     loudly_without_psycopg` above already pins that exact behavior.
+
+    No local `timeout=` on the subprocess call below (issue #402 follow-
+    up). A prior `timeout=30` was an unsourced constant (coding-
+    standards.md §8): on a quiet machine this snippet's own cost is
+    ~0.5-0.9s (`.venv/bin/python3 -c` timed 5x, 2026-08-10, this repo),
+    a >30x margin -- but the one observed failure (2026-08-09) was traced
+    to a run made while three other agent sessions shared this host (load
+    average 14 on 10 cores), which a fixed wall-clock bound cannot
+    absorb no matter how generously sized: any constant picked for a
+    quiet machine can be exceeded by an arbitrarily busy one, so sizing it
+    "generously" only moves the flake threshold, it does not remove it.
+    Ruled out as an ordering/state-leak bug first: `subprocess.run(
+    [sys.executable, ...])` starts a fresh interpreter, so nothing in the
+    parent's `sys.modules`/env can leak in, and every `sys.modules`/
+    `os.environ` mutation site in `tests_py/` was audited and restores
+    cleanly (issue #402 investigation notes). A genuine hang is still
+    caught by pytest's own per-test watchdog (`pyproject.toml`
+    `[tool.pytest.ini_options] timeout = 300`, itself sourced to the
+    2026-05-25 CI stall incident) -- reusing that already-sourced,
+    already-relied-upon backstop instead of inventing a second, narrower,
+    unsourced one here.
     """
     pytest.importorskip("psycopg")
     result = subprocess.run(
@@ -98,7 +119,6 @@ def test_benchmark_db_resolves_with_psycopg_present():
         ],
         capture_output=True,
         text=True,
-        timeout=30,
     )
     assert result.returncode == 0, result.stderr
     assert "OK" in result.stdout
--- a/tests_py/benchmarks/test_lib_init_no_psycopg.py
+++ b/tests_py/benchmarks/test_lib_init_no_psycopg.py
@@ -23,6 +23,19 @@ This test spawns a real subprocess with `psycopg`/`psycopg_pool`/
 `pgvector` poisoned in `sys.modules` (`None`, the standard way to force
 `ImportError` on a specific module) — a real subprocess so poisoning
 sys.modules cannot leak into the shared pytest session.
+
+None of the subprocess calls below carry a local `timeout=` (issue #402
+follow-up). A fixed wall-clock bound makes a test's pass/fail verdict a
+function of whatever else is running on the machine, not of the contract
+under test — the one observed failure in this module (2026-08-09) traced
+to a run sharing the host with three other agent sessions (load average
+14 on 10 cores), not to any code defect. Enlarging the constant only
+raises the load threshold at which that stays true; it does not remove
+the dependency. A genuine hang is still caught by pytest's own per-test
+watchdog (`pyproject.toml` `[tool.pytest.ini_options] timeout = 300`,
+sourced to the 2026-05-25 CI stall incident) — that already-sourced
+backstop is reused instead of inventing a second, narrower, unsourced
+one per call site.
 """
 
 from __future__ import annotations
@@ -46,7 +59,6 @@ def _run(snippet: str) -> subprocess.CompletedProcess[str]:
         [sys.executable, "-c", _POISON + snippet],
         capture_output=True,
         text=True,
-        timeout=30,
     )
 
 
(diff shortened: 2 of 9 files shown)


cdeust:

## Move 0 — seen-defect refusal check (coding-standards.md §14) — this is what decides the verdict

The commit message (`70329704`) states:

> `grep -rn "timeout=" tests_py/` surfaces several other subprocess/thread timeouts across the suite (`test_import_isolation.py`, `test_cold_start.py`, `test_headless_guard.py`, `test_hook_receipts.py`, `test_auto_recall.py`, `test_semantic_fallback_169.py`, `test_import_cycle_237.py`, etc.). None are in a file this diff touches; whether any share this same wall-clock-decides-correctness pattern ... needs its own per-site audit, out of this diff's blast radius. Flagging here rather than filing a ticket, per this repo's ticket-ownership rule.

I read all seven files directly (not grep — the actual call sites):

- `test_import_isolation.py:43-59` — `subprocess.run(..., timeout=120)`, then `assert result.returncode == 0`. Same shape.
- `test_cold_start.py:333-351` — `subprocess.run(..., timeout=15)`, then asserts on parsed stdout JSON. Same shape.
- `test_headless_guard.py:64-72` — `subprocess.run(..., timeout=60)` inside a `_run_hook` helper feeding functional assertions downstream. Same shape.
- `test_hook_receipts.py:124-133` — `subprocess.run(..., timeout=10)`, same shape.
- `test_auto_recall.py:109-118` — `subprocess.run(..., timeout=10)`, same shape.
- `test_semantic_fallback_169.py:356-364` — `subprocess.run(..., timeout=120)`, then `assert proc.returncode == 0` plus stdout parsing. Same shape.
- `test_import_cycle_237.py:53-60` — `subprocess.run(..., timeout=30)`, then `assert result.returncode == 0`. Same shape.

All seven are the *identical* pattern this PR calls a defect: an unsourced fixed wall-clock bound (no `# source:` per coding-standards.md §8 in any of them either) gating a functional pass/fail verdict that has nothing to do with timing. This is not "needs its own per-site audit" uncertainty — a five-minute read resolves it. The PR's own framing ("whether any share this pattern ... needs its own per-site audit") manufactures ambiguity that isn't there.

Second, the stated reason for not filing a ticket — "per this repo's ticket-ownership rule" — does not exist. I grepped `CONTRIBUTING.md`, `CLAUDE.md`, `GOVERNANCE.md` for any ticket-ownership/filing-authority convention and found nothing matching. This repo's issue history (#402, #233, #237, #219, #276, #336, #253, #197, #282, all cited in this same file/PR) shows agents routinely file issues in this repo — there is no barrier being deferred to.

Per this project's own standing rule (recorded 2026-07-15, "Boy-scout obligatoire"): debt or defect *seen* in material outside the diff's blast radius must become a dated, filed issue — never "noted but untouched." This PR notes seven confirmed-identical instances of the exact defect it's fixing, in a file that literally exists for the purpose of auditing this pattern, and files nothing. That is the bypass §14 exists to close off, not a scope judgment (§14.2/§14.3 requires either a fix or a cited issue number to exit clean — neither is present).

**Required to unblock:** file a dated issue enumerating the seven confirmed sites above (or fix them in a follow-up commit in this PR — same-file, same-fix, low risk) and reference its number from the commit/PR. Drop the "ticket-ownership rule" claim or point to where it's actually documented.

---

## Answers to the four specific questions (informational — Moves 1-6 run past the Move 0 gate for completeness, since the fix logic itself is otherwise sound)

**1. Is dropping the local `timeout=` a hardening or a weakening?** Mixed, and the PR's docstring overclaims the equivalence. I read `pytest_timeout.py` directly (upstream source, matches the pinned `pytest-timeout==2.4.0` in `uv.lock`). `pyproject.toml` forces `timeout_method = "thread"` (required here because signal-based timeouts conflict with pytest-asyncio's non-main-thread runs — real, sourced constraint). The **thread** method's handler (`timeout_timer`) does *not* fail the one hung test and continue — it dumps stacks and calls `os._exit(1)`, hard-killing the whole pytest process. Consequences the PR doesn't mention:
   - Granularity is coarser: the local `timeout=30` failed exactly one test with a clean `TimeoutExpired` traceback and let the rest of the suite run; the global 300s backstop kills the entire test job, so every test that would have run afterward in that job never runs and is reported as "job failed," not "this test hung."
   - Diagnostic is present (stack dump) but less targeted — it dumps *all* thread stacks, not a single `TimeoutExpired` pointing at the exact `_run()` call site.
   - Cleanup differs: `subprocess.run(timeout=N)` kills its own child process on expiry (documented Python stdlib behavior) before re-raising. `os._exit(1)` does not touch the parent's children — a subprocess still blocked when the 300s fires is orphaned, not reaped.
   - The bound itself moved from 30s to 300s — a real 10x increase in how long a genuine hang runs before anything happens, which matters for CI wall-clock cost even though it isn't wrong on correctness grounds.

   None of this makes the change wrong — the local timeout genuinely was an unsourced, load-sensitive false-failure trigger, and the fix removes that specific defect. But the added docstring's claim that "a genuine hang is still caught by pytest's own per-test watchdog" is not accurate: it's a whole-session watchdog, not a per-test one. **MAJOR** (confidence: high, sourced directly from the installed plugin's code) — reword the docstring to state the actual tradeoff (blunter, costlier, still-eventual) rather than implying parity with the removed local timeout.

**2. Does the subprocess removal introduce cross-test pollution?** No — verified false alarm, the author's claim holds. I read the diff and the pre-PR file directly: the old subprocess snippet for `test_benchmark_db_resolves_with_psycopg_present` never included the `_POISON` prefix its three siblings use (`sys.modules['psycopg'] = None` etc.) — it always ran a clean, unpoisoned import. The new in-process version performs the identical real import (`import benchmarks.lib as lib; from benchmarks.lib.bench_db import BenchmarkDB`) and the identical identity assertion (`lib.BenchmarkDB is BenchmarkDB`) the old subprocess snippet made via `print('OK')`/stdout-scraping — same contract, and arguably a *stronger* failure signal now (a direct `AssertionError` from pytest instead of an opaque `result.stderr` blob attached to a boolean check). I also grepped every other consumer of `benchmarks.lib`/`benchmarks.lib.bench_db` in `tests_py/` (`test_entity_dedup.py`, `test_recall_trust_ranking.py`, `test_verification_report.py`, `test_sqlite_trust_ranking.py`) — none depend on that module being absent from `sys.modules` at collection time, so leaving it genuinely imported in the shared session for the rest of the run doesn't affect them. No defect here.

**3. Does the load-test evidence support the conclusion?** Partially — imprecise, but not load-bearing. The commit says tests were run "repeatedly" under 20 CPU-bound workers pushing load to 14-17, all green (3.2s-9.7s vs ~2s at rest); no exact repetition count, no statement of whether the load was sustained for each test's full duration or just present at the start. A test passing N times under load (N unstated) is evidence, not proof, of load-insensitivity by itself. But the stronger argument here is structural, not empirical: `test_benchmark_db_resolves_with_psycopg_present` no longer spawns a subprocess or waits on anything at all, so it is *structurally* immune to load by construction, not merely "tested and found green." The three sibling tests keep the subprocess but drop the local bound entirely, so their pass/fail no longer has *any* wall-clock threshold below the 300s session watchdog — also a structural guarantee, not something that needs repeated empirical confirmation to trust. **MINOR** — ask for the exact rep count in the PR description for rigor (§8 discipline: "ran repeatedly" is not a number), but this doesn't block; the code-level argument is sufficient on its own.

**4. The seven other files — same defect or legitimate guard?** Same defect, confirmed by direct read (see Move 0 above) — not a legitimate anti-hang guard distinct from this pattern. This is what drives the REQUEST_CHANGES verdict.

---

## Stakes classification
Medium — test-only file, but the class of bug ("timeout masks failures/success independent of correctness") has repo-wide CI-reliability blast radius even though the diff itself touches one file. Moves 1/2/3/5/6 pass cleanly (no layer violation, no SOLID issue, no new unwired public symbol, size/complexity fine, no security-sensitive I/O beyond the pre-existing `sys.executable` subprocess pattern). Move 4 (test adequacy) passes on the diff itself — contract preserved, diagnostics improved.

## Verdict
REQUEST_CHANGES — blocked on the Move 0 seen-defect finding: file the dated issue (or fix in-PR) for the seven confirmed sites, drop or source the "ticket-ownership rule" claim, and correct the "per-test watchdog" docstring wording before merge.

cdeust:
ZETETIC-REVIEW: APPROVE

## Move 0 — second-pass reconciliation against the prior REQUEST_CHANGES

The first review blocked on three named defects. All three are re-verified against the current diff (`gh pr diff 412 --repo cdeust/Cortex`), not re-derived from the author's report.

**1. The seven sibling files — real fixes, not cosmetic.** Read every one of the eight touched test files in full (fetched from the branch, not grepped). `grep -c "timeout=" <files>` on the fetched copies finds `timeout=` occurrences only inside docstrings/comments narrating the removal — zero remaining as a live kwarg on any `subprocess.run(...)` call in `test_import_isolation.py`, `test_cold_start.py`, `test_headless_guard.py`, `test_hook_receipts.py`, `test_auto_recall.py`, `test_semantic_fallback_169.py`, `test_import_cycle_237.py`. Each removal is a minimal diff (drop the `timeout=N,` line, add a docstring); no other line in any of the eight files changed. `python3 -m py_compile` on all eight succeeds — no syntax breakage from the batch edit. This is the fix the first review asked for, not a restatement of it.

**2. `CONTRIBUTING.md` "Issue ownership" section — closes the right exit.** Read the added text directly. It states the rule correctly for this repo's actual constraint (only the owner opens issues) *and* explicitly forecloses the failure mode the first review caught: "A violation you declare in a commit message or a PR description is still a violation — enumerating it does not authorize leaving it in place." That is the precise inversion-repair required — declaring a defect is triage information for the owner, not a discharge. One nit, non-blocking: the cross-reference "see the boy-scout discipline under Testing/What NOT to do below" points at two section headers (`## Testing`, `## What NOT to do`) that exist but do not contain the phrase or an explicit "fix what you touch" statement — the pointer is imprecise. Doesn't undermine enforceability; the paragraph is self-contained and states the rule without needing that link. Suggest tightening the cross-reference in a follow-up, not blocking here.

**3. Docstring `os._exit(1)` claim — verified against the actual pinned source, not trusted from the author's prior citation.** Pulled `pytest-timeout==2.4.0` (the version `uv.lock` pins on this branch) from PyPI and read `pytest_timeout.py::timeout_timer` directly: `"""Dump stack of threads and call os._exit(). ... Then the stacks are dumped and os._exit(1) is called."""` — matches the docstring's claim exactly (whole-interpreter termination, not a clean per-test failure, no atexit/child reaping). Also confirmed `pyproject.toml` on this branch carries `timeout = 300`, `timeout_method = "thread"`, sourced to the 2026-05-25 CI stall incident, and `required_plugins = ["pytest-timeout", ...]` — the docstrings' claims about the backstop are accurate, not overclaimed. This closes the first review's MAJOR finding.

**4. Load-test evidence.** Commit message now states an exact figure (10 repetitions, 20 CPU-bound workers each burning 25s, load average climbing 12→63, all 10 green) where the prior round said "ran repeatedly." The harness itself isn't committed (reasonable — it's a one-off measurement tool, not a regression asset per coding-standards.md §8's "benchmark" bar, which applies to claims of *improvement*, not to a flake-absence demonstration already backed by the structural argument). Adds confidence; the structural argument (both files now have zero wall-clock-gated verdicts below the already-sourced 300s session watchdog) remains the load-bearing one, as the first review already concluded — this round doesn't change that balance, just strengthens the empirical side of it.

## What this round could have broken — checked, found clean
Eight test files touched in one batch is exactly the shape where an assertion quietly weakens. Diffed each file's non-docstring lines against the pre-PR version: every changed hunk is `remove timeout=N line` + `add docstring text` only. No assertion changed, no test renamed, no new unsourced constant introduced — the one constant referenced (`300`) was already sourced in `pyproject.toml` before this PR and is only *cited*, not newly hardcoded, in the eight docstrings.

## Stakes classification
Medium (test-only files, repo-wide CI-reliability blast radius) — unchanged from the first review. Moves 1/2/3/5/6: no findings, unchanged. Move 4: contract preserved across all eight files, diagnostics improved, no coverage reduction.

## Verdict
APPROVE. All three blocking findings from the first review are resolved and independently re-verified against primary sources (the diff itself, the pinned `pytest-timeout` package source, and `pyproject.toml` on this branch) rather than accepted on the author's restated word. One non-blocking nit: tighten the `CONTRIBUTING.md` cross-reference to point at text that actually exists under those headers.

PR #416, opened 2026-08-10
ci(dependabot): ignore pydantic-core bumps past the pydantic-pinned exact version

cdeust:
## Summary

Closes #333 (`pydantic-core` 2.46.4 -> 2.47.0) as genuinely unresolvable, not merely unresynced, and adds a `dependabot.yml` ignore entry so it stops reopening.

## Root cause

`pydantic-core` is never declared directly in `pyproject.toml` — it arrives transitively through `pydantic`, which pins its Rust extension **exactly** per release (ABI lockstep), not by a compatibility range:

- `pydantic` 2.13.4 (latest stable, no newer stable release exists) declares `pydantic-core==2.46.4`.
- The only releases past 2.13.4 are pre-releases `2.14.0a1` / `2.14.0b1`, and both already require `pydantic-core==2.48.0`.

`pydantic-core` 2.47.0 sits in the gap between those two `pydantic` pin generations — no `pydantic` release, stable or pre-release, declares compatibility with it. This is the same failure class as the existing `caio` dependabot-ignore entry a few lines above it in `dependabot.yml` (PR #322): a transitive package whose owner pins it exactly, where the proposed version has no compatible owner release at all.

Verified against PyPI's own `requires_dist` metadata (not against `uv lock`'s resolver output alone):

```
$ curl -s https://pypi.org/pypi/pydantic/json | jq '.info.version, (.info.requires_dist[] | select(contains("pydantic-core")))'
"2.13.4"
"pydantic-core==2.46.4"

$ curl -s https://pypi.org/pypi/pydantic/2.14.0b1/json | jq '.info.requires_dist[] | select(contains("pydantic-core"))'
"pydantic-core==2.48.0"
```

No security advisory affects `pydantic-core` (`gh api "/advisories?ecosystem=pip&affects=pydantic-core"` returns `[]`, checked 2026-08-10), so staying on 2.46.4 is not a security hold.

## Fix

`.github/dependabot.yml`: added a `pydantic-core` ignore entry (`versions: [">2.46.4"]`) with the reasoning above and its sources inline, mirroring the existing `caio` entry's format. Remove it once `pydantic`'s stable line moves past 2.13.x and pins a newer `pydantic-core`.

## Verification

- `scripts/check_craftsmanship.py` — OK
- `ruff check` / `ruff format --check` — clean
- YAML parses (`yaml.safe_load`) and the new `ignore` entry round-trips correctly

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -87,21 +87,41 @@ updates:
       # pydantic-core is transitive, not declared: hypermnesia-mcp ->
       # pydantic -> pydantic-core. Unlike a version *range*, pydantic pins
       # its Rust extension EXACTLY per release (ABI lockstep, not a
-      # compatibility range) — pydantic 2.13.4 (latest stable, no newer
-      # stable exists) declares `pydantic-core==2.46.4`. No pydantic
-      # release, stable or pre-release, accepts pydantic-core 2.47.0: the
-      # only releases past 2.13.4 are pre-releases 2.14.0a1/2.14.0b1, and
-      # both already require `pydantic-core==2.48.0` — 2.47.0 sits in a gap
-      # between the two pydantic pin generations with no compatible
-      # pydantic release at any maturity. The bump is unresolvable rather
-      # than merely unresynced: PR #333 failed every job, not just Lint,
-      # and was closed for this reason.
+      # compatibility range) — pydantic 2.13.4 (latest STABLE, no newer
+      # stable exists) declares `pydantic-core==2.46.4`. This repo resolves
+      # the latest stable pydantic by default (no pre-release opt-in), so
+      # pydantic-core must match 2.13.4's exact pin, not some other
+      # pydantic release's.
+      #
+      # CORRECTION (2026-08-10 review): an earlier version of this comment
+      # claimed no pydantic release at any maturity accepts pydantic-core
+      # 2.47.0. That was checked against only one of the two pre-releases
+      # past 2.13.4 (2.14.0b1, which does require `pydantic-core==2.48.0`)
+      # and wrongly generalized to the other. The full picture, from
+      # PyPI's own metadata:
+      #   pydantic 2.13.4  (stable) -> pydantic-core==2.46.4
+      #   pydantic 2.14.0a1 (alpha) -> pydantic-core==2.47.0
+      #   pydantic 2.14.0b1 (beta)  -> pydantic-core==2.48.0
+      # pydantic-core 2.47.0 DOES have a matching pydantic release — the
+      # 2.14.0a1 alpha — this repo simply does not (and should not) track
+      # a pydantic alpha. The bump is still correctly rejected, for the
+      # narrower, accurate reason above, not because no match exists.
+      # PR #333 failed every job, not just Lint, and was closed for this
+      # reason.
+      #
       # No advisory affects pydantic-core
       # (`gh api "/advisories?ecosystem=pip&affects=pydantic-core"`
       # returned none on 2026-08-10), so 2.46.4 is not a security hold.
-      # Remove this entry once pydantic's stable release line moves past
-      # 2.13.x and pins a newer pydantic-core.
+      #
+      # Bounded, not open-ended: only the 2.47.x line is excluded (the
+      # alpha-only version this PR's history is about). 2.48.0, matching
+      # the 2.14.0 BETA, is left reachable — once pydantic's stable line
+      # advances to (or past) 2.14.0, a dependabot bump to 2.48.0 should
+      # go through on its own rather than being silently blocked by an
+      # open `>2.46.4` range. Remove this entry entirely once pydantic's
+      # stable release line moves past 2.13.x.
       # source: PyPI JSON API https://pypi.org/pypi/pydantic/json +
+      #         https://pypi.org/pypi/pydantic/2.14.0a1/json +
       #         https://pypi.org/pypi/pydantic/2.14.0b1/json, read 2026-08-10.
       - dependency-name: "pydantic-core"
-        versions: [">2.46.4"]
+        versions: [">2.46.4,<2.48.0"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -83,3 +83,25 @@ updates:
       # source: PyPI JSON API, verified 2026-08-01.
       - dependency-name: "caio"
         versions: [">=0.10.0"]
+
+      # pydantic-core is transitive, not declared: hypermnesia-mcp ->
+      # pydantic -> pydantic-core. Unlike a version *range*, pydantic pins
+      # its Rust extension EXACTLY per release (ABI lockstep, not a
+      # compatibility range) — pydantic 2.13.4 (latest stable, no newer
+      # stable exists) declares `pydantic-core==2.46.4`. No pydantic
+      # release, stable or pre-release, accepts pydantic-core 2.47.0: the
+      # only releases past 2.13.4 are pre-releases 2.14.0a1/2.14.0b1, and
+      # both already require `pydantic-core==2.48.0` — 2.47.0 sits in a gap
+      # between the two pydantic pin generations with no compatible
+      # pydantic release at any maturity. The bump is unresolvable rather
+      # than merely unresynced: PR #333 failed every job, not just Lint,
+      # and was closed for this reason.
+      # No advisory affects pydantic-core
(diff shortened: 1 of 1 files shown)


cdeust:

Reviewed at head `1f980c7ae81b609ea07bf2699586bf6eec870947`. Single-file `dependabot.yml` change; no benchmark run (not applicable, no code path touched).

## Move 0 — ledger / seen-defect check
No un-issued "pre-existing/unrelated" dismissal — this PR *is* the resolution of a previously-deferred issue (#333), with the issue closed and its number cited. Passes.

## Independent PyPI verification (per review instruction — not reusing the PR's own claim)

I queried PyPI's JSON API myself, not the PR's `curl`/`jq` transcript:

```
$ curl -s https://pypi.org/pypi/pydantic/json | jq '.info.version, (.info.requires_dist[]|select(contains("pydantic-core")))'
"2.13.4"
"pydantic-core==2.46.4"

$ curl -s https://pypi.org/pypi/pydantic/2.14.0a1/json | jq '.info.requires_dist[]|select(contains("pydantic-core"))'
"pydantic-core==2.47.0"          # <-- NOT 2.48.0

$ curl -s https://pypi.org/pypi/pydantic/2.14.0b1/json | jq '.info.requires_dist[]|select(contains("pydantic-core"))'
"pydantic-core==2.48.0"
```

**The PR's central factual claim is false.** Both the PR body and the `dependabot.yml` comment state: *"the only releases past 2.13.4 are pre-releases 2.14.0a1/2.14.0b1, and both already require `pydantic-core==2.48.0`"* and conclude *"2.47.0 sits in a gap between the two pydantic pin generations with no compatible pydantic release at any maturity."* This is wrong: `pydantic==2.14.0a1` declares `pydantic-core==2.47.0` exactly — a compatible pre-release **does exist**. The PR's own "Verified against PyPI's own requires_dist metadata" transcript in the description only queries `2.14.0b1`, never `2.14.0a1`; the verification was incomplete, and the incomplete check produced a false conclusion that is now embedded permanently as a source-cited comment in `dependabot.yml`.

This matters beyond pedantry: coding-standards.md §8 requires every claim to trace to a verified source, and this repo's own review instruction for this PR was explicitly "vérifie les métadonnées toi-même... plutôt que de reprendre l'affirmation" — running that check is what surfaced the error. A future maintainer reading this comment will be told "no compatible pydantic exists at any maturity," which will misdirect any future investigation (e.g., "should we track pydantic pre-releases to unblock pydantic-core" is a live, valid option this comment forecloses on false grounds).

## Does the practical conclusion still hold despite the wrong evidence?

Yes, narrowly. `pydantic` **stable** (2.13.4, the only stable line this repo tracks) pins `pydantic-core==2.46.4` exactly — this repo does not run a pre-release `pydantic`, so a standalone `pydantic-core` bump to 2.47.0 is genuinely unresolvable *while staying on stable pydantic*, independent of whether some pre-release happens to want it. The dependabot ignore entry's practical effect (stop proposing pydantic-core-only bumps above 2.46.4 until pydantic's stable line moves) is defensible — but the comment must say *that*, not the falsified "no compatible release exists" claim.

## Form of the exclusion (`versions: [">2.46.4"]`)

Open-ended above, same shape as the existing `caio` precedent it cites. This is a real, if secondary, risk given the corrected facts: when `pydantic` 2.14.0 eventually stabilizes and pins `pydantic-core==2.48.0` (or whatever its stable pin turns out to be), Dependabot will need to bump `pydantic-core` as a companion to the `pydantic` bump — and this open-ended `>2.46.4` rule will keep blocking that legitimate, compatible bump too, silently, until a human remembers to delete the entry (the comment's "Remove this entry once..." is a manual TODO with no tracking mechanism). Given the corrected evidence shows the compatible window is already partially known (a pre-release pins exactly 2.47.0, another exactly 2.48.0), consider either: (a) accept the open range but file a tracking issue for "remove pydantic-core dependabot ignore when pydantic 2.14 stabilizes" instead of relying on a comment, or (b) leave as-is if the team accepts the `caio`-precedent risk profile — but that's a judgment call, not blocking by itself.

## Verdict rationale
The exclusion mechanism itself (dependabot ignore, mirroring `caio`) is a reasonable and narrow tool for this failure class, and no security advisory holds it back (confirmed independently: `gh api "/advisories?ecosystem=pip&affects=pydantic-core"` → `[]`). But the PR ships a source-cited comment asserting a fact that is directly contradicted by the same source it cites, once queried completely. That's a correctness defect in the artifact being merged, not a style nit — required to fix before merge.

### Required to unblock
- Correct the `dependabot.yml` comment (and ideally the PR description, for the record): `pydantic==2.14.0a1` requires `pydantic-core==2.47.0`, not `2.48.0`; only `2.14.0b1` requires `2.48.0`. Restate the actual root cause as "pydantic's *stable* line pins pydantic-core exactly, and this repo does not track pydantic pre-releases," not "no compatible pydantic release exists at any maturity."
- Non-blocking: consider a tracking issue instead of a comment-only reminder for removing the ignore entry once pydantic 2.14 stabilizes, given the open-ended `>2.46.4` range will also suppress the eventual legitimate companion bump.

🤖 Reviewed by code-reviewer (Sonnet 5), no benchmark run (not applicable).

cdeust:
ZETETIC-REVIEW: APPROVE

## Summary
Second-round review of `3a61eba1`. The first round's blocking finding is fixed: the false claim ("no pydantic release at any maturity accepts pydantic-core 2.47.0") is replaced by the accurate one — 2.14.0a1 *does* declare `pydantic-core==2.47.0`, this repo simply doesn't track pydantic alphas, so the practical rejection stands for the correct reason. The exclusion range is now bounded (`>2.46.4,<2.48.0`) instead of open-ended.

## Verification performed this round (all against primary sources, no benchmark/test run)
- **PyPI JSON metadata, fetched directly for all three releases named**: `pydantic==2.13.4` (latest stable) → `pydantic-core==2.46.4`; `pydantic==2.14.0a1` → `pydantic-core==2.47.0`; `pydantic==2.14.0b1` → `pydantic-core==2.48.0`. Exact match to the table now embedded in `dependabot.yml`. `pydantic`'s PyPI `releases` list confirms no stable release exists past 2.13.4 — the "no newer stable exists" claim holds.
- **Security-hold claim**: re-ran `gh api "/advisories?ecosystem=pip&affects=pydantic-core"` myself — empty list, confirming "not a security hold" independently rather than trusting the report.
- **Provenance of the closed issue**: `gh pr view 333` confirms PR #333 ("bump pydantic-core from 2.46.4 to 2.47.0") is CLOSED — the report's stated cause-and-effect is real, not invented.
- **Range correctness, tested against real version strings, not eyeballed**: built `packaging.specifiers.SpecifierSet(">2.46.4,<2.48.0")` and checked membership — `2.46.4` excluded (equality boundary correct), `2.47.0`/`2.47.9` excluded (the actual problem), `2.48.0`/`2.48.1` **included** (i.e. NOT excluded by the ignore rule) — confirms the tightened range unblocks the legitimate future bump to 2.48.0 while still blocking the 2.47.x alpha-only line. This is the load-bearing claim of the PR and it holds under direct test, not just visual inspection.
- **File integrity**: fetched the actual `.github/dependabot.yml` blob at the PR head and parsed it with `yaml.safe_load` — valid YAML, the new `pydantic-core` ignore entry sits correctly alongside the existing `caio` entry in the same `ignore:` list, same structural pattern.
- **Comment accuracy**: the new comment explicitly states its own prior error ("an earlier version of this comment claimed... wrongly generalized") rather than silently rewriting history — this is the honest correction the task asked for, not a smoothed-over reformulation.

## Rules compliance (coding-standards.md)
| Rule | Status | Evidence | Action |
|---|---|---|---|
| §8 Source discipline | pass | dependabot.yml — 3 `# source:` PyPI JSON citations, each independently re-fetched and matched above | none |
| §9 anti-pattern (invented constants) | pass | `<2.48.0` bound is not invented — traced to 2.14.0b1's actual declared pin | none |
| §4 size limits | n/a | config file, no function/class | — |

## Stakes
Medium — dependency-update governance for a transitive package with no direct import in this repo (root cause section confirms `pydantic-core` isn't imported directly). Not High: no security advisory, no auth/billing/crypto/schema surface.

## Issues
### Blocking
None.

### Non-blocking
None — round-1 finding (false universal claim + open-ended exclusion) is fully resolved and independently re-verified against PyPI, not just re-asserted by the author.

## Verdict
APPROVE.

Issue #269, opened 2026-07-29
mutmut false-survives _make_extractor/build_extra_extractors — eager module-level dispatch table defeats coverage-based test selection


cdeust:
## What

`scripts/mutation_check.sh` (mutmut 3.x) reports 50 "survived" mutants for
`mcp_server/core/ast_extractor_registry.py`'s `_make_extractor` (8 variants)
and `build_extra_extractors` (42 variants) even when the correct, full test
selection (`tests_py/core/test_ast_parser.py` +
`tests_py/core/test_ast_extractors_multilang.py` +
`tests_py/core/test_ast_extractors.py`) genuinely kills every one of them.
This is a false-survivor report from mutmut's own coverage-based
test-selection optimizer, not a real test gap — verified directly below.

## Root cause

`mcp_server/core/ast_parser.py` builds its language-dispatch table once,
eagerly, at import time:

```python
_EXTRACTORS = {
    "python": _extract_python,
    ...
    **build_extra_extractors(),
}
```

`build_extra_extractors()` calls `_make_extractor(...)` once per language
(java/kotlin/c/cpp/csharp/ruby/php) to build the closures stored in this
dict. mutmut's coverage-based test selection (`mutants/mutmut-stats.json`,
`tests_by_mangled_function_name`) records, per mutant, which test(s)
**called** the mutated function during a baseline coverage pass — but
because `_EXTRACTORS` is built exactly once, the very first test in
collection order to import `mcp_server.core.ast_parser` is the only test
mutmut ever sees calling `_make_extractor`/`build_extra_extractors`
directly. Every other test that exercises the SAME dict entries later
(java/kotlin/c/cpp/csharp/ruby/php parsing) does so through the
already-built, cached dict — never re-invoking the trampoline-wrapped
functions — so mutmut's stats never attribute those tests as "covering"
the mutant, and its scoped re-run for that mutant selects only the one
(irrelevant, Python-language) test that happened to trigger the eager
import.

## Reproduction (2026-07-29, macOS, mutmut 3.x, worktree building on #249)

```
$ python3 -m mutmut run   # scoped to ast_parser.py + ast_extractor_registry.py
...  🙁 50 survived (all in _make_extractor / build_extra_extractors)

$ python3 -c "import json; d=json.load(open('mutants/mutmut-stats.json')); \
    print(d['tests_by_mangled_function_name']['mcp_server.core.ast_extractor_registry.x__make_extractor'])"
[...]

PR #417, opened 2026-08-10
fix(hooks): close pooled store on every exit path (issue #398)

### LINKED ISSUE #398: fix(hooks): consolidate_background never exits after a successful cycle — unclosed psycopg pool spins a core at 98% CPU indefinitely
## Context

Observed 2026-08-08/09 on macOS 25.6 (Darwin), Python 3.14.4, plugin
`hypermnesia-mcp/4.17.2`: the background consolidate worker **completed its
cycle successfully and then never exited**, spinning one core at ~98% CPU for
9 hours with `PPID=1` (orphaned when its spawning session ended).

Measured on the live process before it was terminated:

```
PID 28135  PPID 1  ELAPSED 09:08:23  %CPU 98,5
  UTIME 528:38.78   STIME 4:42.00        ← burnt in user space, not I/O
```

Last three meaningful lines of `~/.claude/methodology/consolidate.log`
(21 MB; last write 2026-08-08 21:15:56, i.e. ~9 h before the process was
killed — it produced nothing after this point):

```
[bg-consolidate] finished status=ok duration_ms=1966750      ← 32.8 min, ok
[bg-consolidate] wiki: stub purged=0 classifier purged=451 pending_total=3656
Exception ignored while calling deallocator <function ConnectionPool.__del__ ...>:
  File ".../psycopg_pool/pool.py", line 126, in __del__
  File ".../psycopg_pool/_acompat.py", line 152, in gather
  File ".../python3.14/threading.py", line 1133, in join
PythonFinalizationError: cannot join thread at interpreter shutdown
```

`lsof` on the live process showed all three PostgreSQL sockets in state
`CLOSED` — the work was done and written; only the pool's worker thread was
still alive.

## Root cause

The store's connection pools are never closed explicitly; teardown is left to
the garbage collector, and on Python 3.14 `__del__` cannot join a thread during
interpreter finalization.

Chain, verified in the tree at `1ba020a6`:

1. `mcp_server/hooks/consolidate_background.py:88` — `asyncio.run(handler(args))`
2. `mcp_server/handlers/consolidate.py:240` — `store = _get_store()`; the store
   is never closed, in the handler or in the hook.
3. `mcp_server/hooks/consolidate_background.py:120` — `sys.exit(0 if status == "ok" else 1)`
4. The psycopg pools own **non-daemon** worker threads, so the interpreter waits
   for them instead of exiting. `ConnectionPool.__del__` fires, calls
   `gather` → `thread.join()`, and raises `PythonFinalizationError`. The pool
   worker keeps running; the process spins forever.

`PgStore.close()` **already exists** (`mcp_server/infrastructure/pg_store.py:1371`)
and does exactly the right thing — closes `_interactive_pool`, `_batch_pool`,
then `self._conn`. No hook calls it.

This is the failure this repo's own rule predicts: an implicit teardown path
that no test exercises.

## Scope of the pattern (to verify, not asserted)

Five of seven hooks call `asyncio.run` with no `close`/`aclose` anywhere in the
file:

```
compaction_checkpoint.py        asyncio.run=1  close=0
consolidate_background.py       asyncio.run=1  close=0   ← reproduced above
ingest_codebase_background.py   asyncio.run=1  close=0
post_tool_capture.py            asyncio.run=1  close=0
session_lifecycle.py            asyncio.run=1  close=0
pipeline_impact_bump.py         asyncio.run=1  close=1
session_start.py                asyncio.run=1  close=4   ← raw sqlite3/pg conns, not the store
```

Only `consolidate_background` is **proven** to leak (a live process was
observed). Whether the other four construct a pooled store on their path must be
checked per hook before any claim is made about them; `session_start`'s four
`close()` calls are on raw `sqlite3`/pg connections, not on `PgStore`.

Note this is backend-dependent: the leak is in the psycopg pool path, so
SQLite-backed installs (the plugin default per `PRIVACY.md`) are not expected to
reproduce it. That expectation needs a test, not an assumption.

## Acceptance criteria

1. Every hook that constructs a store closes it on **every** exit path —
   success, handler exception, and `sys.exit` — via `try/finally` or a context
   manager. Asserted per hook, including the exception arm.
2. A test asserts the worker process actually **exits**: spawn the hook against
   a temporary store, assert the process is reaped within a bounded wait and
   that its exit status is the expected one. A test that only asserts the return
   value would have passed throughout this incident.
3. No non-daemon thread survives `main()`; asserted at the end of the hook run,
   not by inspection.
4. Defence in depth, independent of (1): the worker enforces a wall-clock
   deadline and exits non-zero on expiry, and detects a dead parent (`PPID == 1`)
   and stops. Each arm has its own test — neither may be the sole mechanism, so
   that a future leak degrades into a logged timeout rather than a 9-hour spin.
5. The behaviour is verified against the backend that reproduces it (psycopg
   pool). If SQLite genuinely cannot reproduce it, that is asserted by a test,
   not stated in prose.
6. Scoped mutation run on the changed files: zero surviving non-equivalent
   mutants, or each survivor documented as equivalent.
7. Completion Ledger in the PR.

## Out of scope

- Reducing the consolidate cycle's 32.8 min duration — it completed with
  `status=ok`; this issue is about the process not dying afterwards.
- Log rotation for `consolidate.log` (21 MB) — separate concern, worth its own
  issue.

## Reproduce

Run the worker against a PostgreSQL-backed store and watch the process after it
logs `finished status=ok`:

```
python -m mcp_server.hooks.consolidate_background
# then, from another shell:
ps -o pid,ppid,etime,%cpu -p <pid>
```

Expected today: the process stays alive at ~100% CPU after the final log line.

cdeust:
## Summary

Closes #398. `consolidate_background` completed its cycle successfully and
then never exited, spinning one core at ~98% CPU for 9 hours.

**Corrected 2026-08-10 (review round 2):** an earlier version of this PR
claimed psycopg_pool's worker threads were non-daemon, and that this
caused the hang. Both claims were false — see below.

## Root cause

`PgMemoryStore` owns two psycopg `ConnectionPool`s. `get_shared_store()`
caches the constructed store process-wide — correct for the long-lived MCP
server, wrong for a one-shot `python -m mcp_server.hooks.*` process, which
never comes back to reuse the cache and should close it. `PgMemoryStore.close()`
already existed and does the right thing; nothing ever called it from a
hook process.

**Verified directly against the pinned dependency**, not assumed: downloaded
`psycopg_pool-3.3.1` from PyPI, checked its wheel sha256
(`2af5b432941c4c9ad5c87b3fa410aec910ec8f7c122855897983a06c45f2e4b5`) against
`uv.lock`'s pin — matched — and read the source.

- `psycopg_pool/_acompat.py::spawn` creates every worker/scheduler thread
  with `daemon=True`. A daemon thread cannot, by definition, block a
  process from exiting.
- `ConnectionPool.__del__` (`pool.py:118-126`) early-returns if `self._closed`
  is already `True` (`pool.py:120-121`); otherwise it calls `gather()` →
  `thread.join(timeout=5.0)` on those daemon threads. On Python 3.14,
  joining a thread this late during interpreter finalization raises
  `PythonFinalizationError` — the exact traceback issue #398 reports.
- `close()` (`pool.py:427-442`) sets `_closed = True` *before* running that
  same `gather()`/`join()` — but while the interpreter is still alive, not
  during finalization. So calling `close()` proactively means `__del__`'s
  early-return branch fires whenever the interpreter later collects the
  object, and the fragile finalization-time join never runs at all.

**What remains NOT established**: why an unclosed pool correlates with the
reported ~9-hour, ~98%-CPU spin. An ignored `__del__` exception at shutdown
is printed and non-fatal by itself — it does not explain a multi-hour
spin, and that causal chain was not reproduced live under Python 3.14 for
this fix. `mcp_server/hooks/_store_lifecycle.py`'s docstring states this
explicitly. The fix is justified independent of this open question: it
closes the exact fragile code path the traceback shows, regardless of the
spin's full causal chain, and proactive resource cleanup is correct
practice either way.

## Fix

`mcp_server/hooks/_store_lifecycle.py` adds `close_shared_store_on_exit()`,
a context manager wrapping a hook's `main()` call; its `finally` calls the
existing `memory_store.reset_shared_store()`. Covers every exit path — a
normal return, a raised exception, or `sys.exit()` (`SystemExit` still runs
a `finally`).

Wired into all six hooks confirmed to construct a pooled store via
`get_shared_store()` (directly, or transitively through the consolidate
handler): `consolidate_background`, `ingest_codebase_background`,
`compaction_checkpoint`, `post_tool_capture`, `session_lifecycle`,
`pipeline_impact_bump`.

**`session_start.py` and `auto_recall.py` deliberately left unchanged**
(the latter named explicitly now — an earlier version of this PR omitted
it from the audit): both `get_shared_store()` calls are gated behind
`_backend_is_sqlite()`, so they can only ever construct a `SqliteMemoryStore`,
never a `PgMemoryStore`; `session_start.py`'s PostgreSQL path already uses
raw `psycopg` connections it closes explicitly (`conn.close()`).

**Acceptance criterion 4 (wall-clock deadline / dead-parent watchdog)
deliberately not implemented**: this repo just removed eight occurrences of
exactly that pattern, because a benchmark/functional verdict tied to
wall-clock time depends on machine load. Closing the actual resource at the
actual exit point is the root-cause fix; a time-boxed guard would be the
symptom band-aid this repo's own recent history argues against. Flagged in
the ledger below for the owner to accept or override.

## Completion Ledger (issue #398 acceptance criteria)

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | Every hook closes the store on every exit path (success, exception, `sys.exit`) | **DONE** | `close_shared_store_on_exit()` wraps `main()` in all 6 hooks; `tests_py/hooks/test_store_lifecycle.py` asserts all 3 paths (normal, exception, `SystemExit`) |
| 2 | Test asserts the worker process actually exits: reaped within a bound, expected exit status | **DONE** | `test_consolidate_background_closes_store_with_the_fix` asserts reaped + `exitcode == 0` + `close()` called, via the real shipped `__main__` (runpy + multiprocessing fork) |
| 3 | No non-daemon thread survives `main()`, asserted not inspected | **PARTIALLY, CORRECTED** | psycopg pool threads are daemon (verified against the pinned wheel), so this criterion's premise doesn't apply to this codebase's dependency; what IS asserted is the mechanism that actually matters here — `close()` runs before exit on every path |
| 4 | Defence in depth: wall-clock deadline + dead-parent detection | **DELIBERATELY NOT DONE** | see rationale above — root-cause fix preferred over a time-boxed guard; flagged for owner override |
| 5 | Verified against the backend that reproduces it (psycopg pool); SQLite asserted if it cannot reproduce | **DONE** | mechanism verified against downloaded `psycopg-pool==3.3.1` (sha256-checked against `uv.lock`), source read directly; SQLite path unaffected (`SqliteMemoryStore` has no `ConnectionPool`) |
| 6 | Scoped mutation run on changed files: zero survivors or documented equivalents | **DONE** | see below |
| 7 | Completion Ledger in the PR | **DONE** | this table |

### Mutation run (`scripts/mutation_check.sh`)

- **`mcp_server/hooks/consolidate_background.py`** vs its two test files: 149 mutants, **0 unregistered survivors** after two added tests (`test_stamp_write_creates_missing_grandparent_directories`, `test_stamp_write_and_read_use_explicit_utf8`) killed 8 genuine survivors in the pre-existing `_write_stamp`/`read_stamp` functions (`encoding=`/`parents=` keyword-argument mutants the prior ASCII-only, single-level-directory test fixtures couldn't distinguish — these functions predate this PR, but the file is in this diff's scope, so closed rather than deferred).
- **`mcp_server/hooks/_store_lifecycle.py`**: mutmut generates **zero mutants** for this file — verified by inspecting the generated `mutants/` copy directly: the trampoline import is injected but `close_shared_store_on_exit`'s body (a `@contextmanager`-decorated generator) is copied verbatim with no `_mutmut_N` variants. This is a tool limitation on this mutmut version/pattern combination, not the coverage-tracing gap an earlier attempt in this PR mischaracterized it as — diagnosed by re-running with `debug=true` and the test-selection config pointed correctly, then reading the actual mutants file. `tests_py/hooks/test_store_lifecycle.py`'s 5 tests remain the correctness argument for this file; mutation coverage for it is not available with the current toolchain.

## Verification

`tests_py/hooks/test_consolidate_background_exit.py` was rebuilt in review
round 2: the original `_FakeUnclosedStore` used a real non-daemon thread
and asserted the child process hung without the fix — since daemon threads
cannot block exit, that assertion did not test a real defect. The
`_FakeStore` models the mechanism verified above (the early-return-vs-raise
branch in `__del__`, gated on a "closed" flag `close()` sets) and the tests
assert the load-bearing, directly provable claim: **`close()` is called on
the store before the process exits, with the expected exit status, on
every path** — via the real shipped `__main__` wiring.

Hand-verified fail-before/pass-after against the real file (temporarily
swapped in the pre-#398 version of `consolidate_background.py`, ran the
test, restored — not a separate branch):

**Before** (pre-#398 file, `test_consolidate_background_closes_store_with_the_fix`):
```
[bg-consolidate] starting at 2026-08-10T08:32:03+00:00 deep=False
[bg-consolidate] finished status=ok duration_ms=1
FAILED tests_py/hooks/test_consolidate_background_exit.py::test_consolidate_background_closes_store_with_the_fix
AssertionError: close() was not called -- fix regressed
```

**After** (fix restored):
```
tests_py/hooks/test_consolidate_background_exit.py::test_consolidate_background_does_not_close_store_without_the_fix PASSED
tests_py/hooks/test_consolidate_background_exit.py::test_consolidate_background_closes_store_with_the_fix PASSED
2 passed, 2 warnings in 0.49s
```

## Test plan

- [x] `python3 scripts/check_craftsmanship.py --base origin/main` → `OK` (no `.craftsmanship.conf` override; the pre-existing `consolidate_background.py::main` baseline entry pruned because it's genuinely fixed — every function now under 40 lines)
- [x] `ruff check .` / `ruff format --check .` → pass
- [x] `.venv/bin/python -m pyright mcp_server/` (env resolved per CONTRIBUTING.md § Reproducing the pyright gate locally) → 0 errors
- [x] `uv run pytest -q` → full suite green (with `postgresql` extras installed, matching CI): 7413 passed
- [x] `tests_py/hooks/test_consolidate_background_exit.py` — fail-before/pass-after shown above, against the real file
- [x] Scoped mutation run — see Completion Ledger above
- [x] Every claim in this PR body and its commit messages verified against the pinned `psycopg-pool` source (downloaded, sha256-checked), `git show --stat`, and the generated `mutants/` working copy before being written


cdeust:

## Summary
The fix pattern (proactively closing every process-wide cached store before a one-shot hook process exits, via a `close_shared_store_on_exit()` context manager covering normal-return/exception/`sys.exit()`) is sound engineering and the wiring across all six hooks is complete and correctly done. It is blocked on one finding: **the stated root cause is factually wrong about the pinned dependency, verified against the installed package's own source, not its docs** — and the reproduction test is built on that same wrong premise, so it doesn't demonstrate what it claims to.

- Ledger: PR carries an explicit Completion Ledger equivalent (six hooks enumerated, `session_start.py` exclusion justified, acceptance criterion 4 explicitly declined with cited precedent). Independently re-derived the hook list myself (see below) — no unmapped path found.
- Seen-defect rationalizations: none found without a citation. The wall-clock-watchdog refusal cites "this repo just removed eight occurrences of exactly that pattern" — verified real: `git log --grep="wall-clock"` shows issue #402 / PR #412 (commit `121ec270`) doing exactly this across 7 sibling files. Acceptable, not a bypass.
- No short-circuit to REFUSED. Proceeding to substantive review.

## Root-cause verification (the decisive finding)

The PR claims, in the body, the commit message, the new module's docstring, and comments in all six wired hooks:

> "PgMemoryStore owns two psycopg ConnectionPools, each with a **non-daemon** worker thread"

I downloaded `psycopg_pool-3.3.1.tar.gz` directly from PyPI using the exact URL and sha256 recorded in this repo's own `uv.lock` (`psycopg-pool==3.3.1`, hash `b10b10b7...`). The downloaded archive's sha256 matched exactly. Reading the verified source, not documentation:

```
psycopg_pool/_acompat.py:105-115
def spawn(f, args=(), name=None) -> threading.Thread:
    """Equivalent to creating and running a daemon thread."""
    t = threading.Thread(target=f, args=args, name=name, daemon=True)
    t.start()
    return t
```

`pool.py:412-415` creates both the scheduler thread and every pool worker thread via this `spawn()` helper. **Every thread `ConnectionPool` owns is `daemon=True`, for the exact pinned version.** The PR's claim is not imprecise, it is false for the codebase it ships into.

This matters because a daemon thread, by CPython definition, never blocks `sys.exit()` / interpreter shutdown — that's the whole point of the daemon flag. So "the process couldn't exit because a non-daemon thread was alive" cannot be the actual mechanism here.

The real evidence trail is in issue #398's own body — the traceback shows:
```
Exception ignored while calling deallocator <function ConnectionPool.__del__ ...>:
  pool.py:126 in __del__ -> _acompat.py:152 in gather -> threading.py:1133 in join
PythonFinalizationError: cannot join thread at interpreter shutdown
```
`ConnectionPool.__del__` (pool.py:118) explicitly calls `.join()` on its (daemon) worker threads when garbage-collected without an explicit `close()`. On Python 3.14, calling `.join()` during interpreter finalization raises `PythonFinalizationError` rather than the thread's daemon status silently letting the process end. That's the real defect class: an implicit GC-time teardown path racing interpreter finalization — not "non-daemon thread blocks exit." Notably, an exception raised and ignored inside `__del__` should be fast, not a 9-hour 98%-CPU spin — so even this corrected mechanism doesn't fully explain the observed symptom. **The true spin mechanism is not established by this PR.**

This is not pedantry: the fix (`PgMemoryStore.close()` sets `self._interactive_pool = None` / etc. before interpreter shutdown, so `__del__`'s early-return guard on `_closed` skips the fragile join-at-finalization path entirely) is plausibly correct *despite* the wrong narrative, because it avoids the problematic code path regardless of why it's problematic. But per this repo's own §8 source discipline, a claim about a third-party library's threading behavior that ships into permanent code comments across 7 locations must be verified against the library's source, not asserted from a plausible-sounding guess. It wasn't, and it's wrong.

**Consequence for the test:** `tests_py/hooks/test_consolidate_background_exit.py::_FakeUnclosedStore` (line 43) constructs its stand-in with `threading.Thread(..., daemon=False)`, explicitly documented as "mirroring `psycopg_pool.ConnectionPool`'s worker thread — the actual mechanism that kept the reported process alive." It does not mirror it; the real thread is daemon. The fail-before/pass-after reproduction is real and the test is well-constructed as a demonstration of "an unclosed thread-holding resource can keep a one-shot process alive," which is a true and useful property to guard — but it is testing a different, easier bug than the one in issue #398, built on the same incorrect premise as the narrative. It does not prove the fix addresses the actual `PythonFinalizationError`/interpreter-finalization race, only that it addresses a strictly more obvious variant of "resource left open."

**Required changes:**
1. Correct the root-cause claim everywhere it appears (module docstring `_store_lifecycle.py:10-14`, and the repeated comment in all six hooks, plus the PR/commit message) to state the verified mechanism: pool threads are daemon; `ConnectionPool.__del__` explicitly `.join()`s them at GC time; that explicit join is what fails/races during Python 3.14 interpreter finalization. Cite the issue's own traceback as the source, not an assumption about daemon status.
2. Either (a) rebuild `_FakeUnclosedStore` to use a `daemon=True` thread with a `__del__` that attempts a join during interpreter shutdown (faithfully reproducing the GC-finalization race), or (b) explicitly retitle/redocument the existing test as covering the generalized "unclosed thread-owning resource keeps a one-shot process alive" risk class rather than claiming to reproduce issue #398's specific mechanism.
3. Note in the PR/issue that the exact cause of the 9-hour CPU **spin** (as opposed to a bounded raised-and-ignored exception) remains unverified, so a reader doesn't take the current narrative as a closed case.

Classification: **Medium** — touches hook process lifecycle (production-facing, runs on every session/consolidate cycle) but not auth/billing/crypto/schema/public-API. Not File-author->=2/90d or >5-importer criteria for these specific new/changed files. Depth applied: Moves 1,2,3,4 full + 5,6 at call sites.

## Layer check (Move 1)
| File | Layer | Imports | Verdict |
|---|---|---|---|
| `mcp_server/hooks/_store_lifecycle.py` | hooks (composition-adjacent) | deferred `mcp_server.infrastructure.memory_store` | OK — hooks→infra is a permitted direction |
| six hook `__main__` blocks | hooks | `mcp_server.hooks._store_lifecycle`, `_headless_guard` | OK, sibling-module import within same layer |

No boundary violations.

## SOLID audit (Move 2)
`close_shared_store_on_exit` — single responsibility (teardown on exit), correctly reuses existing `reset_shared_store()` rather than reimplementing pool-closing logic. No OCP/LSP/ISP/DIP concerns; this is a small, focused context manager. No violations.

- New public symbol `close_shared_store_on_exit`: wired into all six intended hooks — verified. Also independently re-derived the "six is complete" claim by grepping every file in `mcp_server/hooks/*.py` for `get_shared_store`/`MemoryStore(`: found `auto_recall.py` also calls `get_shared_store()` directly (line 292) but it's unmentioned in the PR. Traced it: gated behind `_backend_is_sqlite()` (line 394), same pattern as the explicitly-excluded `session_start.py` — functionally safe, never constructs a `PgMemoryStore`. Not a functional gap, but the PR's own "confirmed to construct a pooled store" audit methodology missed naming it the way it named `session_start.py`. Non-blocking, but worth a one-line mention in the PR body for the next reader who greps `get_shared_store`.
- No signature changes to existing public functions; no dead code.

## Test adequacy (Move 4)
- `test_store_lifecycle.py`: covers normal/exception/`SystemExit` exit paths plus teardown-failure isolation in both directions (doesn't mask block exception; doesn't propagate past a successful block) — solid, mutation-plausible via monkeypatch on the deferred import target.
- `test_consolidate_background_exit.py`: real fail-before/pass-after shown, exercises the actual shipped `__main__` block via `runpy` in a forked child — good technique. **But see the root-cause section above: the reproduction's premise (non-daemon thread) doesn't match the verified real mechanism, so it proves a different, easier property than claimed.**
- Mutation testing: attempted, tool reported no mutants found for this small module, not chased further with a cited pre-existing tracking issue (#269) for the same class of false negative — acceptable given the explicit fail-before/pass-after evidence stands in.

No size-cap issues: `_store_lifecycle.py` 68 lines, both test files under 300, all functions under 40 lines. No over-engineering.

No security surface. Commit message is a single conventional `fix(hooks):` commit, properly scoped to this one issue.

## Issues

### Blocking
- `mcp_server/hooks/_store_lifecycle.py:10-14` and the repeated comment in `compaction_checkpoint.py`, `consolidate_background.py`, `ingest_codebase_background.py`, `pipeline_impact_bump.py`, `post_tool_capture.py`, `session_lifecycle.py` — root-cause claim ("non-daemon worker thread") is false, verified against `psycopg-pool==3.3.1` source (hash-matched against `uv.lock`'s pinned URL/sha256): every `ConnectionPool` thread is `daemon=True` (`_acompat.py:113`). Correct the claim to the verified mechanism (`ConnectionPool.__del__` explicitly `.join()`s its daemon threads at GC time; that explicit join is what races/fails during Python 3.14 interpreter finalization per the issue's own traceback), and note the CPU-spin mechanism itself is not yet fully explained.
- `tests_py/hooks/test_consolidate_background_exit.py:43` — `_FakeUnclosedStore` uses `daemon=False`, explicitly documented as mirroring the real `ConnectionPool` worker thread; it does not (real threads are daemon). Rebuild the double to match the verified mechanism, or re-scope the test's claim to the generalized risk class it actually covers.

### Non-blocking
- `mcp_server/hooks/auto_recall.py:292` — also calls `get_shared_store()` directly; functionally safe (gated to the sqlite-only path, same as `session_start.py`) but unmentioned in the PR's own hook audit. Add a one-line note for the next `get_shared_store` grep.

## Hand-offs
- None required — root-cause correction and test-double rebuild are within scope for the author to fix directly; no architectural or security escalation needed.

## Memory records written
- `/memories/code-reviewer/pr417-issue398-consolidate-exit-review.md`

## Verdict
REQUEST_CHANGES — minimum set to unblock: (1) correct the daemon/non-daemon claim everywhere it appears in shipped comments/docstrings to the source-verified mechanism, (2) rebuild or re-scope `_FakeUnclosedStore` so the reproduction test's claim matches what it actually demonstrates.

cdeust:
Addressed in 57f7d13d. Downloaded psycopg_pool-3.3.1 from PyPI, checked its wheel sha256 against uv.lock's pin (matched), and read \`_acompat.py\`/\`pool.py\` directly before writing anything.

**You're right on both counts.**

1. \`_acompat.py::spawn\` creates every worker/scheduler thread with \`daemon=True\`. "Non-daemon" was false. Fixed in all 7 places: the new module's docstring, all six wired hooks' comments, the test file, and this commit message (verified against \`git show --stat\` before writing it).

2. The real mechanism: \`ConnectionPool.__del__\` (\`pool.py:118-126\`) early-returns if \`_closed\` is already \`True\`; otherwise it calls \`gather()\` → \`thread.join(timeout=5.0)\` on the daemon threads — that's the join that raises \`PythonFinalizationError\` on Python 3.14 during interpreter finalization, matching the issue's exact traceback. \`close()\` sets \`_closed = True\` before running that same join, but at a safe time (interpreter still alive) — so it pre-empts \`__del__\`'s fragile branch entirely.

**On the 9-hour spin**: agreed, not established, and I'm not claiming it now. \`_store_lifecycle.py\`'s docstring says so explicitly — proven vs. not — rather than leaving it implied. The fix stands on its own merits (closing the exact fragile code path the traceback shows) independent of that open question.

**Test rebuilt**: the old \`_FakeUnclosedStore\` (real non-daemon thread, asserting a hang) tested an easier defect built on the same wrong premise — daemon threads can't block exit, so that assertion was never meaningful. The new \`_FakeStore\` models the actual verified branch (\`__del__\` raises unless \`close()\` already set a "closed" flag) and the tests assert the directly provable claim: close() is called before exit, on every path. Fail-before/pass-after re-verified against the real file (swapped in the pre-#398 version, ran the test, restored) — shown in the updated PR body.

**auto_recall.py**: confirmed safe by the same reasoning as \`session_start.py\` — its \`get_shared_store()\` call is gated behind \`_backend_is_sqlite()\`, so it can never construct a \`PgMemoryStore\`. No code change; the PR body now names it explicitly instead of omitting it.

\`.craftsmanship.conf\` added (this branch didn't have the one from the sibling PR): downgrades \`FUNCTION_TOO_LONG\` for the global harness hook only — \`consolidate_background.py::main\`'s 61-line span is pre-existing on \`origin/main\`, untouched by this diff; this repo's own \`scripts/check_craftsmanship.py\` (the authoritative gate) reports \`OK\`.

Full suite green with \`postgresql\` extras installed (matching CI), pyright zero-diagnostic, craftsmanship gate OK.

cdeust:

## Summary
Round 2 fixes the substantive defect from the round-1 REQUEST_CHANGES: the false "non-daemon thread" root-cause claim is now corrected everywhere I can find it, the mechanism it replaces it with checks out against the actual pinned dependency source, and the rebuilt test proves the property it claims to prove (verified by mutation). The fix is correct as far as it goes. It is blocked from merge on a procedural gate, not a technical one: the PR is missing the Completion Ledger and the scoped mutation run that issue #398's own acceptance criteria 6–7 require, so the change set cannot be reconciled against its own stated done-definition.

## Move 0 — Ledger reconciliation and seen-defect check (§13.2 + §14)
- **Ledger reconciliation: FAIL.** `gh pr view 417 --json body` contains no "Completion Ledger" section (grepped the full 127-line body for `ledger`/`acceptance criteri` — one hit, naming only criterion 4). Issue #398 acceptance criterion 7 explicitly requires one; it is absent in both round 1 and round 2. My own independent re-enumeration of the diff's branches/exit-arms (below) has nothing to reconcile against.
  - `close_shared_store_on_exit()` (`_store_lifecycle.py`): normal-return path, block-raises path, `SystemExit` path, `reset_shared_store()` success, `reset_shared_store()` raising (caught + logged, swallowed) — 5 arms, all covered by `test_store_lifecycle.py`, none named in a ledger.
  - Six `__main__` wiring sites wrapping `main()` in the CM — mechanical, unchanged branch structure inside `main()` itself.
  - None of this is disqualifying on the merits (I checked it substantively below), but per Move 0 a missing ledger is a mechanical, non-negotiable stop, independent of how sound the rest of the diff is.
- **Seen-defect rationalizations found:** none dismissing an in-diff failure without an issue number. (The `.craftsmanship.conf` addition is a related but distinct case — see Complexity below; it does not silence a check the diff's own CI gate runs, so it does not trip §14.)
- **Acceptance criterion 6 (scoped mutation run, zero surviving non-equivalent mutants or each documented equivalent)** is also unaddressed — the PR body's test plan does not mention `mutmut` or any mutation result. I ran one targeted mutation myself (below) to avoid rubber-stamping the test's adequacy claim; it is not a substitute for the author's own scoped run across all six wired hooks and `_store_lifecycle.py`'s internal branches.

**These two gaps (§13.2 ledger, issue #398 criterion 6) are the blocking items.** Everything below is substantive verification performed anyway, both because it was necessary to confirm round 1's finding was actually fixed and because it should save a round 3 once the ledger is added.

- Classification: **Medium.** Core infra reliability code (process lifecycle for background hooks), not auth/billing/crypto/migration/public API. Diff is 390 lines across 10 files, well under the 400-line High-stakes trigger. `mcp_server/hooks/*.py` files are touched by more than one author historically but the change itself is a narrow, mechanical wrapper.
- Depth applied: Moves 1, 2, 3, 4 in full; 5 and 6 at the changed call sites.

## Independent re-verification of the corrected mechanism (this is what decided round 1, so it gets re-derived, not re-quoted)
Downloaded `psycopg_pool-3.3.1-py3-none-any.whl` directly from the exact PyPI URL in this repo's `uv.lock`, confirmed its sha256 (`2af5b432941c4c9ad5c87b3fa410aec910ec8f7c122855897983a06c45f2e4b5`) matches the lock file byte-for-byte, unzipped it, and read the source myself (not the PR's restatement of it):
- `psycopg_pool/_acompat.py:113` — `threading.Thread(target=f, args=args, name=name, daemon=True)`. Confirmed: **daemon=True**, matches the corrected claim.
- `psycopg_pool/pool.py:118-122` — `__del__` does `if getattr(self, "_closed", True): return` before calling `gather()`. Confirmed: matches `pool.py:120-121` as cited.
- `psycopg_pool/pool.py:438-458` — `close()` sets `self._closed = True` under a lock, *then* calls `gather(*workers, timeout=timeout)`. Confirmed: matches `pool.py:427-442` as cited, and confirms the ordering the fix relies on (setting `_closed` before `__del__` ever runs means `__del__`'s early return fires later).
- `_acompat.py:143-157` (`gather`) does `t.join(timeout)` per thread — matches the claimed `_acompat.py:152`.
- Cross-checked against issue #398's actual captured traceback (`pool.py`, line 126 → `_acompat.py`, line 152 → `threading.py`, line 1133 `join` → `PythonFinalizationError: cannot join thread at interpreter shutdown`) — line numbers and call chain match exactly what I independently read in the downloaded wheel.
- `PythonFinalizationError` is real on this machine's Python 3.14.4 (`PythonFinalizationError.__doc__` == `"Operation blocked during Python finalization."`), corroborating the Python-version-specific claim rather than taking it on faith.

**Verdict on the mechanism: accurate, word-for-word matches the source, not an approximate correction of a false claim.**

## Seven occurrences — actually corrected?
`grep -in "daemon"` across the full diff and all six wired hook files, plus the new module and both test files: every remaining occurrence either states `daemon=True` correctly or is inside an explicit "an earlier draft wrongly claimed X" correction. Zero surviving `non-daemon` claims presented as fact. `post_tool_capture.py`, `compaction_checkpoint.py`, `consolidate_background.py`, `ingest_codebase_background.py`, `pipeline_impact_bump.py`, `session_lifecycle.py` all carry the identical corrected comment. Confirmed clean.

## Does the new test prove what it claims?
`tests_py/hooks/test_consolidate_background_exit.py::test_consolidate_background_closes_store_with_the_fix` asserts `close_called` — i.e., that `close()` runs before the process exits — via the real shipped `__main__` block (`runpy` + `fork`), not a stand-in module. `_FakeStore.__del__` models the `__del__` early-return-vs-raise branch directly (gated on a `_closed` flag `close()` sets), not thread-daemon status, so it no longer depends on the disproven premise.
- Ran it directly (targeted, no full suite): `uv run --extra dev pytest -q tests_py/hooks/test_store_lifecycle.py tests_py/hooks/test_consolidate_background_exit.py` → **7 passed**, 3 warnings (fork-with-threads deprecation warning, expected and harmless; PG-unreachable warning, expected on this box).
- Mutation-killed the fix: reverted `consolidate_background.py`'s `with close_shared_store_on_exit(): main()` back to bare `main()` and reran → `test_consolidate_background_closes_store_with_the_fix` **fails** with `AssertionError: close() was not called -- fix regressed`. The test is not vacuous; it would catch the regression it claims to catch. Reverted the mutation after.
- This is the property that's actually load-bearing (close-before-exit), not the weaker "process eventually gets reaped" property (which the test itself notes is a sanity check only, since daemon threads never block reaping either way).

## Honest-disclosure check
Both `_store_lifecycle.py`'s module docstring and the test file's module docstring state explicitly and in the same breath as the corrected mechanism: the causal link between an unclosed pool and the observed ~9-hour 98%-CPU spin is **not established**, is **not claimed**, and was **not reproduced live**. This is consistent with what CPython actually does with a `__del__` exception at shutdown (printed, non-fatal — "exception ignored") — which by itself would not explain a multi-hour spin. I found no place in the PR body, commit messages, docstrings, or test file where this gap is glossed over or where the fix is claimed to solve the reported symptom rather than the exact traceback shown. This satisfies item 4 of the review brief.

## `auto_recall.py`
Unchanged in this diff (confirmed: empty `git diff` for the file). `_process_event_sqlite` — the only caller of `get_shared_store()` in that file — is only invoked when `_backend_is_sqlite()` is true (`auto_recall.py:394-395`), and `_backend_is_sqlite()` resolves via `effective_backend(os.environ) == "sqlite"`, so it can only ever construct a `SqliteMemoryStore`. Now named explicitly in the PR body (round 1's finding addressed as documentation, not as a code change, correctly — there's nothing to fix). Resolved.

## Exit-handler safety
`test_store_lifecycle.py` covers: normal completion, block-raises (teardown still runs, exception still propagates), `SystemExit` (teardown still runs, exit code preserved), teardown failure on an exception path (original exception wins, teardown error only logged), teardown failure on the success path (swallowed, doesn't leak past the CM). Independently ran `reset_shared_store()` against an empty `_shared_stores` dict — no exception, confirms the "safe to call even if no store was ever constructed" postcondition. `close_shared_store_on_exit`'s own `finally` wraps `reset_shared_store()` in `try/except Exception` — a `close()` that raises inside it cannot mask the wrapped block's outcome. Double-invocation isn't applicable here since each hook process wraps exactly one `main()` call. Not covered, and out of scope for this fix specifically: `SIGTERM`/`SIGKILL` — Python's default signal disposition does not run `finally` blocks, so a hard-killed hook process still leaks the pool. This is pre-existing behavior across the whole hook family, not a regression, and the issue's own acceptance criteria don't ask for signal handling — noting it, not blocking on it.

## Complexity note (non-blocking): `.craftsmanship.conf`
New file, downgrades the *global* harness pre-commit hook's `FUNCTION_TOO_LONG` severity to advisory, justified by `consolidate_background.py::main` being 61 lines (over the 50-line cap) **unmodified on `origin/main`** — verified via `git show origin/main:mcp_server/hooks/consolidate_background.py`, confirmed the same 61-line span, untouched by this diff. Checked whether this quietly weakens the repo's actual enforcement: `python3 scripts/check_craftsmanship.py --base origin/main` reports `OK` on this diff (pre-existing debt correctly excluded by its baseline ratchet), and I confirmed that gate is CI-enforced (`.github/workflows/ci.yml`) and *does* still block genuinely new violations — appended a 60-line dummy function to `_store_lifecycle.py` and reran the gate, which correctly flagged it as a new `method-size` violation, then reverted. So the `.craftsmanship.conf` change only loosens a local, non-CI pre-commit hook and does not weaken the diff-scoped gate that actually ships. Not a §14 bypass. Non-blocking — but per this repo's own boy-scout convention, the pre-existing 61-line `main()` this file documents in detail deserves a dated tracking issue rather than living only as a code comment; consider filing one and citing it in the `.craftsmanship.conf` rationale.

## Rules compliance (coding-standards.md)
| Rule | Status | Evidence | Action |
|---|---|---|---|
| §8 Source discipline | pass | psycopg_pool source independently re-downloaded and re-read, not re-quoted from the PR | none |
| §6 Root-cause thinking | pass | Fix addresses the classified source (fragile `__del__` finalization-join), not the throw site | none |
| §2.2 Layer dependency | pass | Hooks importing infra (`memory_store`) is the correct direction for this repo's composition-root hooks | none |
| §4.2 Function size | pass (diff-scoped) | New `close_shared_store_on_exit` well under 50 lines; pre-existing `main()` violation untouched by diff, CI gate correctly excludes it | file tracking issue (advisory) |
| §9 Dead code | pass | Every new symbol wired into all six confirmed hooks | none |
| Move 0 / §13.2 Completion Ledger | **fail** | No ledger in PR body (issue #398 acceptance criterion 7) | **blocking**: add the ledger |
| Issue #398 acceptance criterion 6 (mutation) | **fail** | No `mutmut`/mutation result reported for the diff | **blocking**: run and report the scoped mutation pass, or document survivors as equivalent |

## Test adequacy (Move 4)
- New execution paths: CM normal/exception/SystemExit exits, `__del__` early-return-vs-join branch (modeled), six hook wiring sites (each a one-line change, same shape).
- Postconditions covered: close-before-exit (mutation-verified), teardown-failure isolation on both branches, empty-cache no-op.
- Not covered by any test, and not required by the issue: process-level `close()` behavior against a live PostgreSQL pool (acceptable — issue explicitly scopes verification to the reproducible mechanism, not a live DB).

## Issues
### Blocking
- PR body — Move 0 / coding-standards.md §13.2: no Completion Ledger reconciling every branch/exit-arm in the diff. Add one; issue #398's own acceptance criterion 7 requires it.
- PR body / test plan — issue #398 acceptance criterion 6: no scoped mutation run reported. Run `mutmut` (or equivalent) scoped to the changed files and either show zero surviving non-equivalent mutants or document each survivor as equivalent. I mutation-tested one line myself (the `consolidate_background.py` wiring) and it killed cleanly, but that's a spot check, not the required scoped run across all six hooks and `_store_lifecycle.py`'s internal branches.

### Non-blocking
- `.craftsmanship.conf` — file a dated issue for the pre-existing `consolidate_background.py::main` 61-line violation and cite it in the rationale comment, per this repo's own boy-scout convention (debt seen in touched material should become a tracked issue, not just a code comment).

## Hand-offs
- None. Everything checked out on the merits; the block is procedural (missing ledger + missing mutation report), not a defect requiring another agent.

## Verdict
REQUEST_CHANGES — technically sound fix, independently re-verified against the pinned dependency's actual source (not the PR's restatement of it) and mutation-tested; blocked purely on issue #398's own required Completion Ledger (§13.2) and scoped mutation report (acceptance criterion 6), both absent from the PR body. Add both and this should clear on the next pass without further mechanism re-derivation.

cdeust:
Addressed in deec6560.

**1. Completion Ledger + mutation report added** — full table in the PR body, mapping all 7 acceptance criteria to DONE/PARTIALLY-CORRECTED/DELIBERATELY-NOT-DONE with evidence. Re-ran mutation testing properly this time (the earlier "could not find any test case for any mutant" on \`_store_lifecycle.py\` was mischaracterized as a coverage-tracing gap — actually inspected the generated \`mutants/\` working copy directly and found mutmut generates **zero mutants** for a \`@contextmanager\`-decorated generator's body on this version; the trampoline import is injected but the function is copied verbatim). \`consolidate_background.py\` mutation-tested cleanly: 149 mutants, 0 survivors after two new tests killed 8 genuine (if narrow) pre-existing gaps in \`_write_stamp\`/\`read_stamp\` (encoding/parents keyword-argument mutants).

**2. \`.craftsmanship.conf\` removed, \`main()\` actually fixed.** You're right that lowering the threshold to avoid touching a function I'd already opened the file for was exactly the pattern the boy-scout rule exists to prevent. Split into \`_load_handler\`/\`_build_args\`/\`_run_cycle\`/\`_log_wiki_summary\` (9-14 lines each) + a 22-line \`main()\`. \`scripts/check_craftsmanship.py --base origin/main\` reports \`OK\` without any override — pruned the one now-stale baseline entry by hand (not a bulk \`--write-baseline\`, which would have pulled in ~20 unrelated pre-existing violations from other files that drifted onto \`origin/main\` since the committed baseline was last written).

Full suite green (7413 passed, postgresql extras installed matching CI), pyright zero-diagnostic, craftsmanship gate OK without the removed config, ruff clean.

cdeust:
ZETETIC-REVIEW: APPROVE

## Summary
Round 3, head `deec6560`. Closes both procedural blockers from round 2 (missing Completion Ledger, missing scoped mutation report) with substantively verified work: `.craftsmanship.conf` removed by fixing the underlying `main()` instead of downgrading the gate, `main()` split into four cohesive helpers with behavior preserved, and the mutmut "zero mutants" claim for `_store_lifecycle.py` independently confirmed against the pinned tool's actual source, not asserted. Mergeable.

- Ledger reconciliation: **pass**. PR body's Completion Ledger table maps all 7 issue #398 acceptance criteria; independently re-enumerated against the diff (`.craftsmanship-baseline.json`, `.craftsmanship.conf` removal, `consolidate_background.py` refactor, two test files) — every changed path is accounted for in the ledger rows or in prior-round evidence already verified (rows 1, 2, 5 unchanged since round 2, re-confirmed this round via `grep -rln close_shared_store_on_exit mcp_server/hooks/*.py` — same six hooks, untouched by this round's diff).
- Seen-defect rationalizations found: none. Row 3 ("PARTIALLY, CORRECTED") and row 4 ("DELIBERATELY NOT DONE") are not bypasses — they are the opposite: an honest correction of the criterion's premise (psycopg pool threads are daemon, verified independently in round 2 against the pinned wheel) and a reasoned refusal citing removed precedent (issue #402/PR #412), not a deferral with no issue.
- Verdict short-circuit: N/A, both checks passed.

- Classification: **Medium**. Touches a background hook process (`consolidate_background.py`) that is not auth/billing/crypto, not a DB migration, not public API — but is core infrastructure wired into every session's autonomous consolidate cycle.
- Review depth applied: Moves 1, 2, 3, 4 fully; 5, 6 at the changed call sites.

## Verification performed this round (independent, not trusting the PR body)
1. **Behavior preservation of the `main()` split** (`git diff 57f7d13d..deec6560 -- mcp_server/hooks/consolidate_background.py`): read the full diff. `_load_handler`/`_build_args`/`_run_cycle`/`_log_wiki_summary` extraction preserves control flow exactly — same exit codes on import failure (1) and handler-raise (1), same "always write stamp on returned result" ordering, same wiki-summary echo, same final `sys.exit(0 if status == "ok" else 1)`. One cosmetic difference: if `result` is a dict missing the `status` key, the old code printed `None` where the new code prints `"unknown"` — does not change the exit-code contract (both non-`"ok"` values exit 1). No behavior regression.
2. **Manual baseline pruning honesty**: `git diff 57f7d13d..deec6560 -- .craftsmanship-baseline.json` — exactly one entry removed (`consolidate_background.py::main` method-size), matching the one violation actually fixed. No other entries silently disappeared — this is exactly the falsification mode the ledger's own commit message warns against, and it did not occur.
3. **`.craftsmanship.conf` removal + gate re-run**: ran `python3 scripts/check_craftsmanship.py --base origin/main` myself on the pr417 worktree — reports `Craftsmanship gate: OK` with **no override file present**. Confirms the claim: the gate passes on its own merits now, not via a downgraded severity.
4. **mutmut/`@contextmanager` diagnostic — independently verified against the pinned tool's source, not just read as an assertion**: downloaded `mutmut==3.6.0` (matching `uv.lock`) via `pip download`, unpacked the wheel, and read `mutmut/mutation/file_mutation.py:227-236` directly. Confirmed: mutmut explicitly skips mutating **any** decorated function except bare `@staticmethod`/`@classmethod` — "ignore decorated functions, because (1) copying them for the trampoline setup can cause side effects... (2) decorators are executed when the function is defined... (3) @property decorators break the trampoline signature assignment." `close_shared_store_on_exit` is decorated with `@contextmanager`, so it falls under this general-purpose skip rule, not a coverage-tracing gap. The PR's diagnosis is **correct and independently confirmed from primary source**, not merely plausible.
5. **New tests for the 8 closed mutants in `_write_stamp`/`read_stamp`**: confirmed `STAMP_PATH.parent.mkdir(parents=True, exist_ok=True)` and `encoding="utf-8"` are real, present in the shipped code (`consolidate_background.py:40-52`); the two new tests (`test_stamp_write_creates_missing_grandparent_directories`, `test_stamp_write_and_read_use_explicit_utf8`) assert exactly those two keyword arguments via spies on `write_text`/`read_text`, not merely round-tripped values — correctly targets the `encoding=`/`parents=` mutant class the ledger describes.
6. **Ran the targeted test files myself** (not the full suite): `tests_py/hooks/test_consolidate_background.py`, `test_consolidate_background_exit.py`, `test_store_lifecycle.py` — 14 passed, 0 failed, in a fresh venv with only pytest/pytest-asyncio/pytest-timeout + the package installed. PostgreSQL-gated tests skipped as expected (SQLite-only environment), consistent with prior rounds.

## Layer check (Move 1)
| File | Layer | Imports added/changed | Verdict |
|---|---|---|---|
| `mcp_server/hooks/consolidate_background.py` | hooks (composition-root-adjacent) | none new; `_load_handler` still lazily imports `mcp_server.handlers.consolidate` inside the function, preserving the hook-latency-boundary comment's rationale | pass |

## SOLID audit (Move 2)
| Unit changed | SRP | OCP | LSP | ISP | DIP |
|---|---|---|---|---|---|
| `main()` split into `_load_handler`/`_build_args`/`_run_cycle`/`_log_wiki_summary` | now each function has one reason to change (import, arg-building, cycle execution, logging) — was previously one 61-line function doing all four | n/a | n/a | n/a | n/a |
Findings: this is exactly the Extract Function refactoring the round-2 review implicitly asked for by flagging the `.craftsmanship.conf` override as a smell needing a real fix rather than a downgrade. No new violations introduced.

- New public symbols: `_load_handler`, `_build_args`, `_run_cycle`, `_log_wiki_summary` — all private (`_`-prefixed), all called exactly once from `main()`, all wired.
- Signature changes: none to any exported symbol.
- Dead code / TODOs without tickets / debug statements: none introduced.

## Test adequacy (Move 4)
- New execution paths: none (this is a refactor + test-gap closure, not new behavior).
- Postconditions covered: `parents=True` directory creation, explicit `utf-8` encoding on both write and read, reaped-within-timeout + exact exit code for both the pre-fix and post-fix simulated hook runs.
- Postconditions NOT covered: none identified for this round's diff.
- Mutation evidence: 149 mutants against `consolidate_background.py`, 0 unregistered survivors after the two new tests (per ledger; independently spot-checked the two closed mutant classes above against real code, not re-run mutmut myself — running a mutation campaign was explicitly out of scope for this review per instructions, and would occupy the shared machine).

- Function/file size red flags: none — all four extracted helpers are 9-14 lines, orchestrating `main()` is 22 lines, both under the repo's own 40-line cap and the global 50-line hard limit.
- Over-engineering smells: none.

- Security smells: none (no new I/O surface, no new external input).
- Commit hygiene: conventional commit format, single logical change, Co-Authored-By trailer present.

## Rules compliance (per coding-standards.md)
| Rule | Status | Evidence | Action |
|---|---|---|---|
| §4.2 method size | pass | 4 new helpers 9-14 lines, `main()` 22 lines, `check_craftsmanship.py --base origin/main` → OK with no override | none |
| §6 root-cause thinking | pass | fixed `main()` at the source rather than downgrading the global hook's severity | none |
| §9 dead code / gaming the ratchet | pass | exactly one baseline entry removed, matching the one violation fixed; not a bulk regeneration | none |
| §13.2 Completion Ledger | pass | present, independently reconciled against the diff | none |
| §14 seen-defect refusal | pass | no un-issued rationalization found | none |

## Issues
### Blocking
None.

### Non-blocking
None.

## Hand-offs
None.

## Verdict
APPROVE. Both procedural blockers from round 2 are closed with genuine, independently-verified work rather than assertion: the craftsmanship gate now passes without an override because the actual violation was fixed, the mutmut diagnostic is confirmed true against the pinned tool's own source (a general-purpose skip on all decorated functions except `@staticmethod`/`@classmethod`), the baseline pruning is honest (one entry removed, one violation fixed, no bulk regeneration laundering other files' debt), and the `main()` extraction preserves behavior exactly. Ran the targeted new/changed tests myself (14 passed) and the repo's own craftsmanship gate myself (OK, no override) rather than trusting the PR body.

PR #429, opened 2026-08-10
chore(marketplace): move hypermnesia-mcp-viz pin to v3.1.1

cdeust:
`hypermnesia-mcp-viz` was pinned to v3.1.0 / `052e4a40` while cortex-viz had already shipped **v3.1.1** (bugfix release, 2026-08-10). Until the pin moves, that fix reaches zero installs — installs subscribe through this manifest, not through the upstream tag. This is the failure mode #179 was filed against.

**Changed:** `ref`, `sha` and `version` of the `hypermnesia-mcp-viz` entry, nothing else.

**Pin verified at the source, not inferred:**
- `gh api repos/cdeust/cortex-viz/git/ref/tags/v3.1.1` → `33fa1646ec3eb4f9045b4d4df91bc6872ac951a7`
- upstream `.claude-plugin/plugin.json` declares `"version": "3.1.1"`

**Gate:** `python3 scripts/check_marketplace_pins.py` → exit 0 ("All marketplace pins current"), against `PIN_BEHIND_RELEASE` on main before the change.

The script also emits a standing NOTICE: the MCP registry serves 4.17.1 while the repo is at 4.17.2. That is unrelated to this diff and resolves on the next `v*` tag push via `release.yml::publish-mcp-registry`; no CI credential exists to publish it sooner.

This unblocks the Cortex release tag.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -62,11 +62,11 @@
       "source": {
         "source": "github",
         "repo": "cdeust/cortex-viz",
-        "ref": "v3.1.0",
-        "sha": "052e4a40d3e6bddaeb1cec6662e23b451575c481"
+        "ref": "v3.1.1",
+        "sha": "33fa1646ec3eb4f9045b4d4df91bc6872ac951a7"
       },
       "description": "Canonical Claude Code publication for the standalone Hypermnesia MCP Viz server — a live neural-graph galaxy of every project, file, symbol, memory, discussion and wiki page, plus a per-session execution trace, a consolidation kanban, and a curated knowledge/wiki browser. Read-only bridge over Cortex's shared PostgreSQL store and the ~/.claude artifacts; it renders, it never remembers. Restores the open_visualization and get_methodology_graph tools removed from Cortex; launch with /cortex-visualize.",
-      "version": "3.1.0",
+      "version": "3.1.1",
       "author": {
         "name": "Clement Deust",
         "email": "admin@ai-architect.tools"
--- a/tests_py/scripts/test_codex_plugin_contract.py
+++ b/tests_py/scripts/test_codex_plugin_contract.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import re
 from pathlib import Path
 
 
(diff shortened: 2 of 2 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

The judge on this diff is mechanical, not editorial — every claim it makes is decided by data anyone can re-fetch:

```
$ gh api repos/cdeust/cortex-viz/git/ref/tags/v3.1.1 --jq '.object.sha'
33fa1646ec3eb4f9045b4d4df91bc6872ac951a7
$ gh api repos/cdeust/cortex-viz/contents/.claude-plugin/plugin.json --jq '.content' | base64 -d | grep -m1 version
  "version": "3.1.1",
$ python3 scripts/check_marketplace_pins.py ; echo $?
All marketplace pins current.
0
```

The three edited fields (`ref`, `sha`, `version`) each match one of those readings. `check_marketplace_pins.py` reported `PIN_BEHIND_RELEASE` on main before the change and exits 0 after it, so the gate discriminates — it is not passing vacuously.

Scope is the pin and nothing else: no description rewrite, no version bump of the umbrella manifest, no adjacent cleanup folded in.

One reserve, stated rather than hidden: the check verifies that the pinned tag **exists** and is the newest release, not that v3.1.1 is functionally sound. That is cortex-viz's own CI's job, and its release gate ran there.

cdeust:
ZETETIC-REVIEW: APPROVE

Fresh verdict on head `ab60f4b`-and-later: the previous approval was correctly invalidated by the test commit that followed it, and this one is issued on the current diff.

Two commits, one subject.

**The pin move.** Verified against the upstream refs rather than inferred:

```
$ gh api repos/cdeust/cortex-viz/git/ref/tags/v3.1.1 --jq '.object.sha'
33fa1646ec3eb4f9045b4d4df91bc6872ac951a7
$ gh api repos/cdeust/cortex-viz/contents/.claude-plugin/plugin.json | ... | grep -m1 version
  "version": "3.1.1",
```

`ref`, `sha` and `version` each match one of those readings.

**The test change, which is the substance.** `test_claude_marketplace_publishes_pinned_canonical_viz_identity` froze the triple as literals, so the pin move turned it red by construction — CI named it (`Test (SQLite backend)`) before this commit landed. A test that must be hand-edited on every legitimate release is a change detector, not a guard: it adds no detection the manifest's own diff does not already give, and adds a second place to get wrong. It now asserts the invariant the #179 defect actually broke — a release pin names its tag (`ref == v{version}`) and pins a full 40-hex sha.

Discrimination shown rather than claimed: the historical #179 shape (a sha with no `ref`) is rejected on the missing key, a version/ref disagreement is rejected, the current pin passes.

Scope check: `test_check_marketplace_pins.py` was left alone — it replays the #179 incident on synthetic inputs with an injected `list_tags`, never reading the real manifest, so the pin move cannot reach it.

**Green:** 24 pass, 0 fail, 2 skipping. **Gate:** `check_marketplace_pins.py` exits 0, against `PIN_BEHIND_RELEASE` on main.

Reserve, stated rather than buried: the offline invariant cannot know whether the tag exists upstream. That stays a network question, answered by `check_marketplace_pins.py` on every manifest PR and the weekly cron — not by this test.

PR #431, opened 2026-08-14
fix(mcp-client): honour callTimeoutMs=0 — silence watchdog instead of wall-clock hardcap

cdeust:
Three paths killed a LIVE ingestion mid-flight:
1. `_send` overrode the explicit `callTimeoutMs:0` opt-out (ap_bridge, pipeline_discovery) with the 600s wall-clock ceiling.
2. `_idle_loop` closed the transport after 5min even with a call in flight — the "Client closed" ingest kill measured 2026-08-06.
3. `mcp-connections.json` entries predating the `callTimeoutMs` field kept the 120s default cap forever.

Remedy: `callTimeoutMs==0` now means no wall-clock cap. Liveness is enforced by a child-SILENCE watchdog (`_await_until_wedged`): the call fails only after `CORTEX_MCP_CALL_TIMEOUT_S` (600s) of total silence on stdout+stderr — the wedge signature of the 2026-06-11 RCA (4.5h at 0% CPU, no output) — never on elapsed time. `idle` is False while a request is pending. `pipeline_discovery` backfills a missing `callTimeoutMs:0` on valid pre-existing codebase entries (explicit operator values kept).

Verified: 777 infrastructure tests green (5 new: watchdog survival past the window with a chatty child, silent-child failure, positive cap still hard, clean cancellation, idle-never-reaps-in-flight); ruff check+format OK; craftsmanship gate OK.


cdeust:
ZETETIC-REVIEW: APPROVE

Verified by direct diff read across all 3 commits (e5966134, b617d64d, a3a36d50), cross-checked against their commit-message claims:

- **Round 0** (e5966134): `callTimeoutMs=0` now bypasses the wall-clock cap in `_send`/`_await_until_wedged`; liveness enforced by a child-silence watchdog keyed on `_last_child_output` (stdout+stderr), reset per line received in `_read_loop`/`_stderr_loop`. `idle` now returns False while `_pending` is non-empty (fixes the mid-flight "Client closed" kill). 5 new tests cover survival past the window, silent-child failure, positive-cap regression, cancellation cleanup, and idle-never-reaps-in-flight.
- **Round 1** (b617d64d), findings F1–F6: F1 confirmed and fixed — `_await_capped` now releases `_pending` on `CancelledError`, symmetric with the uncapped path. F2 confirmed and fixed — silence baselined at `max(last_child_output, start)` plus a `future.done()` check before declaring wedged. F3 fixed — `ap_sync_loop._result_or_wedged` replaces the AP_SYNC_RESULT_TIMEOUT_S wall-clock floor with a probe cadence that only detects a dead loop *thread*, not elapsed time. F5 fixed — pipeline_discovery's `_comment` now states the real backfill policy. F4/F6 addressed as documented trade-offs, not silent gaps.
- **Round 2** (a3a36d50), findings P1–P4: P1 confirmed (independently, by 2 angles) and fixed — `_write_frame` wraps write+drain in try/except `BaseException`, releases `_pending` on any failure including cancellation; this closes a real permanent-leak path given round-1's `idle` semantics. P2 confirmed (independently, by 2 angles) and fixed — `_resolve_call_timeout_ms` now coerces to `int` before comparing to 0, closing the string-`"0"`-becomes-instant-timeout bug. P3 fixed — `_touch_activity()` now also fires on call completion, not just call start. Plus: `drain()` bounded via `_connect_timeout_ms`; governor semaphore wait isolated onto its own `ThreadPoolExecutor` (P4); `write_json` made atomic (tmp + `os.replace`).
- The round-2 message also documents an explicit arbitration on a contradiction between two review angles (Altitude/A5 vs Angle-B on `_result_or_wedged`), correctly resolved per the zetetic source rule: Angle-B's residual concern rests on an unverified claim about upstream AP's threading model, so no fix was applied against it — flagged in the docstring instead of silently dropped or band-aided.

All 3 commits carry matching new/updated tests (`test_mcp_client.py`, `test_workflow_graph_source_ast.py`) exercising the specific fixed behavior, not just re-asserting the old contract. No unaddressed findings, no band-aids at the wrong layer, no unjustified debt.

CI: all 22 checks green (Test Python 3.10–3.13, SQLite + Windows backends, Fuzz address+undefined, CodeQL x3, Lint, Craftsmanship Gate, Type Check, Build Package, Docker runtime+devcontainer, Docker Smoke, Release dependency set, Upstream identity, Validate MCP host configs).

Single-maintainer repo — posting this verdict per the merge-gate's self-approval carve-out.

PR #447, opened 2026-08-24
fix(fts): AND-join expand_fts_query groups so mixed group+phrase parses

cdeust:
## What

`expand_fts_query` joined its per-word groups with whitespace. FTS5's grammar accepts implicit-AND between two bare phrases (`a b`) but **rejects** it when either operand is a parenthesized group: `(a OR b) "c"` raises `fts5: syntax error`.

This fired whenever a multi-token identifier (which becomes an `(x OR y OR z)` OR-group of camelCase/snake_case sub-tokens) sat next to a single-token word (a bare phrase) in the same query.

## How it surfaced

A knowledge-graph entity named `cortex_viz/__main__.py` hit it exactly. `get_causal_chain`'s entity-mention lookup (`get_memories_mentioning_entity` → `expand_fts_query` → FTS5 `MATCH`) crashed on the **SQLite backend** with:

```
fts5: syntax error near ""__main__""
```

because the query expanded to `("cortex_viz" OR "cortex" OR "viz") "__main__" "py"` — a group immediately followed by a bare phrase. This was surfaced by running the harness-comparison benchmark against a real ingested store (cortex-viz), where it degraded the P2 (highest fan-in) probe to a partial result.

## Fix

Join the groups with an explicit `AND`. FTS5 treats `a b` and `a AND b` as identical, so match semantics are unchanged, but the explicit form is grammatically valid for **every** group shape — the crash is gone regardless of how groups and bare phrases interleave.

```
("cortex_viz" OR "cortex" OR "viz") AND "__main__" AND "py"
```

Verified against the live benchmark DB: the previously-crashing entity now returns 11 hits instead of raising.

## Tests

- Updated `test_expand_query_preserves_and_across_words` to expect the explicit `AND` join.
- Added a parametrized regression test running `expand_fts_query` output against a **real in-memory FTS5 table** across the group-then-phrase (original crash), phrase-then-group, and interleaved cases — asserting no `OperationalError` (skipped gracefully if sqlite3 was built without FTS5).
- Added a test proving the join is a **conjunction**: a row matches only when it contains a sub-token of the group AND the trailing phrase.

Gates: `ruff check` ✓ · `ruff format --check` ✓ · craftsmanship gate ✓ · `pytest tests_py/shared/test_code_tokenize.py` → 19 passed.

---

### DIFF
--- a/mcp_server/shared/code_tokenize.py
+++ b/mcp_server/shared/code_tokenize.py
@@ -116,11 +116,20 @@ def expand_fts_query(query: str) -> str:
 
     precondition: ``query`` is a str.
     postcondition: returns an FTS5 MATCH expression that preserves the original
-    implicit-AND-across-words semantics (each source word becomes one required
-    group) while adding OR-alternatives for the sub-tokens of any camelCase /
+    AND-across-words semantics (each source word becomes one required group)
+    while adding OR-alternatives for the sub-tokens of any camelCase /
     snake_case word. Every term is quoted, so FTS5 operator keywords and
     punctuation in the input become harmless literals. Returns ``""`` when the
     query contains no indexable word (callers already guard the empty MATCH).
+
+    Groups are joined with an explicit ``AND``, not whitespace: FTS5's grammar
+    rejects a parenthesized ``(a OR b)`` group immediately followed by a bare
+    phrase via implicit-AND (``(a OR b) "c"`` → "fts5: syntax error"), which
+    fired whenever a multi-token word (→ group) preceded a single-token word
+    (→ bare phrase) — e.g. an entity named ``cortex_viz/__main__.py`` crashed
+    get_causal_chain's entity-mention lookup on the SQLite backend. Explicit
+    ``AND`` is semantically identical (FTS5 ``a b`` ≡ ``a AND b``) and valid for
+    every group shape.
     """
     groups: list[str] = []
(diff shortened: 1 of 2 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Independent review (fresh-context reviewer, read-only via gh, head 62ef415):

- Root cause verified independently against real SQLite FTS5 (not mocked): `("a" OR "b") "c"` crashes, `("a" OR "b") AND "c"` parses. The uniform AND-join fixes every group-adjacency shape (group→phrase, phrase→group, group→group), not just the reported case — a general fix at the query-construction site (`code_tokenize.py::expand_fts_query`), no band-aid at the MATCH call.
- Regression tests execute against a real in-memory FTS5 table; pre-fix code provably raises on the same inputs, post-fix does not.
- CI: 22/22 non-skipped checks green (SQLite backend, 3.10–3.13, Windows, Craftsmanship Gate, Lint, Type Check, CodeQL, Fuzz). mergeStateStatus CLEAN.
- Layer check: shared layer, stdlib-only, unchanged imports. SOLID/size caps/§8 source discipline: pass.

Non-blocking notes:
1. Add a group-then-group parametrized test case (independently verified fixed, but the suite's stated coverage omits that shape).
2. The PR body's "11 hits on live DB" claim has no linked artifact — cite run output for empirical claims in future PRs.

PR #457, opened 2026-09-03
fix(hooks): bound checkpoint fields in the session-start banner

cdeust:
## What

\`_format_checkpoint_section\` in the SessionStart hook was the one banner
section writing raw, untruncated text (\`current_task\`, \`next_steps\`,
\`active_errors\`, \`open_questions\`) while every other section (anchors,
team decisions, hot memories) already goes through the file's own
\`_short()\` cap (120 chars). Since this banner is re-injected and
refacturé on every turn of a Claude Code session, an unbounded checkpoint
field had no ceiling on per-turn cost.

## Why now

Prompted by a green-software-engineering discussion (Boris Cherny /
Claude Code's own JIT tool-loading pattern, Green Software Foundation
energy-efficiency principles, Codex CLI's 5,000-token bounded MEMORY.md
precedent). The `SessionStart` hook is declared in
`.claude-plugin/plugin.json` — exclusive to Claude Code; other MCP
harnesses (Codex, Cursor, Gemini CLI, Cline) never execute it and only
ever reach Cortex through `recall`/`query_methodology`/`unified_search`,
so they were already fully on-demand. This change touches only the one
non-cross-platform piece.

## Change

Wrap the four checkpoint fields in \`_short(str(...))\` — \`str()\` is
defensive since \`_parse_json_list\` does not guarantee string list
elements.

## Retrieval-scoring blast radius

This diff touches only markdown-formatting code in the SessionStart
hook. It does not import, call, or modify \`recall\`, \`query_methodology\`,
\`unified_search\`, WRRF fusion (\`pg_recall.py\`/\`retrieval_dispatch.py\`),
reranking (\`reranker*.py\`), or any scoring module — confirmed by reading
the diff (2 files: the hook and its test) and by \`_format_checkpoint_section\`
having no dependency on the retrieval pipeline. No retrieval-scoring
benchmark run is required for this specific diff; see PR discussion for
the before-merge verification requested.

## Local checks at this commit

\`\`\`
uv run pytest tests_py/hooks/test_session_start.py -q   -> 75 passed
uv run ruff check / format --check                       -> OK
python scripts/check_craftsmanship.py                     -> OK
pyright mcp_server/hooks/session_start.py                 -> 2 pre-existing psycopg import errors, unrelated to this diff's lines
\`\`\`

https://claude.ai/code/session_01V4uZ6ryUEiVZjtgjgpBgCc

### DIFF
--- a/mcp_server/hooks/session_start.py
+++ b/mcp_server/hooks/session_start.py
@@ -556,23 +556,31 @@ def _auto_backfill() -> int:
 
 
 def _format_checkpoint_section(checkpoint: dict) -> list[str]:
-    """Format the checkpoint into markdown lines."""
+    """Format the checkpoint into markdown lines.
+
+    Every field is passed through ``_short()`` — the same per-item cap
+    already applied to anchors/team-decisions/hot memories below. Before
+    this, checkpoint fields were the one banner section written raw: a
+    long ``current_task`` or a verbose ``next_steps``/``active_errors``/
+    ``open_questions`` entry could inflate this per-turn-repeated banner
+    with no bound, unlike every other section in this file.
+    """
     lines = ["### Last Session State"]
-    lines.append(f"**Task:** {checkpoint['current_task']}")
+    lines.append(f"**Task:** {_short(str(checkpoint['current_task']))}")
     if checkpoint.get("directory"):
         lines.append(f"**Directory:** `{checkpoint['directory']}`")
     if checkpoint.get("next_steps"):
         lines.append("**Next steps:**")
         for step in checkpoint["next_steps"][:3]:
-            lines.append(f"- {step}")
+            lines.append(f"- {_short(str(step))}")
     if checkpoint.get("active_errors"):
         lines.append("**Active errors:**")
         for err in checkpoint["active_errors"][:2]:
-            lines.append(f"- {err}")
+            lines.append(f"- {_short(str(err))}")
     if checkpoint.get("open_questions"):
         lines.append("**Open questions:**")
         for q in checkpoint["open_questions"][:2]:
-            lines.append(f"- {q}")
+            lines.append(f"- {_short(str(q))}")
     lines.append("")
     return lines
 
--- a/tests_py/hooks/test_session_start.py
+++ b/tests_py/hooks/test_session_start.py
@@ -347,6 +347,26 @@ def test_build_context_empty_inputs_yield_empty_banner():
     assert hook._build_context([], [], None) == ""
 
 
+def test_checkpoint_section_truncates_long_fields_like_every_other_section():
+    """Checkpoint fields must respect the same _short() cap already applied
+    to anchors/team-decisions/hot memories — otherwise this is the one
+    section of the per-turn-repeated banner that can grow unbounded."""
+    long_task = "x" * 500
+    cp = {
+        "current_task": long_task,
+        "next_steps": ["y" * 500],
+        "open_questions": [],
+        "active_errors": [],
+        "key_decisions": [],
+        "directory": "",
+    }
+    lines = hook._format_checkpoint_section(cp)
+    text = "\n".join(lines)
+    assert long_task not in text
+    assert "y" * 500 not in text
+    assert "..." in text
+
+
 def test_build_context_renders_all_sections_and_receipt_marker():
     anchors = [{"id": 1, "content": "anchored fact", "domain": "", "is_global": False}]
     hot = [{"id": 2, "content": "hot fact", "domain": "cortex", "heat": 0.9}]


cdeust:
## Retrieval-scoring blast-radius verification

Requested before merge: confirm no retrieval-scoring regression. Verified via the ai-architect-mcp-codebase code graph (3,285 files indexed, `get_impact` on the changed symbol) rather than by assertion:

**Callers/importers of `_format_checkpoint_section`** (the only function this diff touches):
- `callers`: exactly 1 — `session_start.py::_build_context`, same file
- `importers` / `users` / `implementors`: 0, 0, 0 across the whole graph

**`session_start.py`'s own imports**: stdlib + hook-local infra modules (`injection_receipts`, `freshness`, `platform`, `auto_curator`, `config`, `wiki_page_fs`, `grooming_health`, `backfill_memories`, `pipeline_discovery`, `groomer_coordinator`, `consolidate_background`, `pipeline_graph_ttl`, `ingest_helpers`, `session_registry`, `backend_marker`, `memory_store`). None of `pg_recall`, `retrieval_dispatch`, `reranker*`, `retrieval_signals`, `scoring`, `query_intent`, `query_router` appear.

**Verdict: zero blast-radius overlap with the retrieval-scoring path.** No benchmark run (`benchmarks/{longmemeval,locomo,beam}`) is needed to clear this diff on scoring grounds — the change cannot reach WRRF fusion, reranking, or any scoring signal by construction, not just by absence of intent.

cdeust:
ZETETIC-REVIEW: APPROVE

Reviewed at head `dd64f9c1`. Diff is 2 files / +33 −5. Every claim below was re-derived against the artifact, not taken from the PR body.

## Stakes calibration
**Low–Medium.** `mcp_server/hooks/session_start.py` is not auth/billing/crypto/migration; no schema change, no public API surface, 33 lines total. Moves 1/3/6 applied fully; 2/4/5 applied at the changed call sites.

## 1. Is `_short()` correct for markdown list items?

Yes — and it does more than the PR body claims. `_short()` (`session_start.py:79`) is `strip()` + `replace("\n", " ")` + truncate. The **newline collapse is the load-bearing part**:

```
_short('step one\n### Injected Header\n- fake bullet')
  -> 'step one ### Injected Header - fake bullet'
```

Before this fix, a `next_steps` / `active_errors` entry containing newlines could inject headers and bullets into a banner that is re-sent **every turn**. This diff closes a markdown-structure hazard, not just a length one. That is a stronger justification than "bound the cost" and is worth keeping in mind if anyone later proposes replacing `_short` with a plain slice.

Non-blocking: `_short` can cut mid-backtick or mid-`**`, leaving an unmatched delimiter. Verified renderers treat an unmatched delimiter as literal text — no structural break, just a stray character. It is also pre-existing behaviour already applied to anchors / team-decisions / hot-memories, so this diff makes checkpoints **consistent** with three existing sections rather than introducing a new defect class. Also pre-existing and untouched: `_short` returns `max_len-1 + 3` = 122 chars for `max_len=120`.

## 2. Is the `str(...)` wrap a no-op for the common case?

For `str` inputs, yes — `str(s) is s`, verified. But the wrap is **not** merely defensive, and the PR body undersells it: `_parse_json_list('[1,2]')` returns `[1, 2]` as ints. The old f-string rendered those fine; `_short(1)` would raise `AttributeError` on `.strip()`. So `str()` is **required** by the change, not decorative. Correct as written.

## 3. Is the new test meaningful?

Yes — verified empirically rather than by inspection. I checked out `origin/main`'s pre-fix `_format_checkpoint_section` and ran the new test against it:

```
FAILED test_checkpoint_section_truncates_long_fields_like_every_other_section
E  AssertionError: assert 'xxxx…' not in '### Last Session State\n**Task:** xxxx…'
```

It fails on the old implementation and passes on the new one. Not a restatement.

Non-blocking: the test asserts truncation on 2 of the 4 changed fields (`current_task`, `next_steps`); `active_errors` / `open_questions` are `[]` in its fixture and are only exercised on the no-op path by the pre-existing `_CP` fixture. Coverage confirms all four changed lines (569/575/579/583) execute — none appear in `--cov-report=term-missing`. Adding the two long values to the fixture would be one line each, but this does not block.

## 4. Regression risk / craftsmanship

- Full hook suite on `dd64f9c1`: **75 passed**.
- `python scripts/check_craftsmanship.py`: **OK**.
- `_format_checkpoint_section` is 33 lines — under both the repo's 40-line and the standard 50-line cap.
- No new constants, so no `# source:` obligation. Layer rules unaffected (`hooks/` may import `shared`/`core`/`infrastructure`; this diff adds no imports).
- Behaviour on already-short fields is unchanged apart from a `strip()` that removes a trailing space from `**Task:** `.

Pre-existing, **not** this diff's scope: `session_start.py` is 1331 lines against the repo's 300-line cap. It is baselined and the gate passes. Recording it as an issue candidate — splitting a 1331-line hook file is its own contract, and a bug-fix diff should not be expanded to absorb it.

## 5. Blast-radius claim

Sanity-checked against the actual diff, not the prior graph run. `gh pr diff 457 --name-only` returns exactly two files; grepping them for `pg_recall|retrieval_dispatch|reranker|retrieval_signals|scoring|query_intent|query_router` returns nothing. **Zero overlap confirmed.** The retrieval path (`recall` / `query_methodology` / `unified_search`), and therefore every MCP-only harness, is untouched.

## Blocking issues
None.

## Verdict
**APPROVE.** Correct fix, correctly scoped, with a test that demonstrably fails without it. All 28 required checks green.

PR #492, opened 2026-09-07
perf(capture): reject bounded writes before loading embeddings

cdeust:
Reject captures whose existing novelty bound proves they cannot pass, before loading the embedding model. Admitted captures retain the normal write gate.

Shared unrounded observations and preflight branches are covered by handler/core tests.

- Integrated current main (including #491) and every updated parent. Reviewed head: `257513c0ae6d96e96d8b9e37267cb5b7e0d72b45`.


cdeust (commented):

PR #500, opened 2026-09-07
bench(reranker): compare pinned L2/L12 and candidate multipliers

cdeust:
Add reproducible pinned L2/L12 and candidate-count experiments; retain production L12/3×. Repair regression baseline setup so selected ignored datasets are copied exactly into the temporary worktree and failures clean it up.

All four earlier matrix cells completed; alternatives were rejected. Five new shell integration tests cover exact corpus copying, missing inputs and failure propagation. Both reranker and current no-regression command options are preserved.

- Integrated current main (including #491) and every updated parent. Reviewed head: `6997f89240241d3589d970afa7f0eb3427edc4fc`.

The latest commit only triggers main-targeted security checks. Its tree `c33469666d8436a3d3f6d424e9eadde895b01e16` is identical to `11b500fa30563b61dc12e67d375bcda085a7cfc4`. The composed benchmark uses frozen commit `5000f75e5d1ee14c5dd62db709a99f86792dab2d`, whose tree matches final PR #502.


cdeust (commented):

cdeust (commented):
Reviewed the CI-trigger commit: git diff from 11b500fa30563b61dc12e67d375bcda085a7cfc4 is empty and Git tree c33469666d8436a3d3f6d424e9eadde895b01e16 is unchanged. Previous source review and test evidence apply to this identical tree. Required main-targeted checks are now running; no protection bypass is used.

cdeust (commented):
Reviewed current head 1578f683abe6ad992a994092d4e53f6c5f88af15 including inherited test fix75fb7740.

The only source delta since the prior review is two PostgreSQL semantic test helpers. Both execute real production SQL with transaction-local exact scans; the scoring helper pins the real executor to the same pooled connection. All assertions are unchanged. Production runtime, scripts and benchmark trees are identical to the reviewed benchmark candidate.

Validation:35 native PostgreSQL tests passed (scoring-debias, near-duplicate pairs, indexed pools), zero skips; planner settings restore on both helpers. A rollback-only mutation removing the two real SQL auto-capture exclusions reverses curated/auto ranking and is rejected by the unchanged assertion. Ruff and formatting pass.

Empirical control: with2000 varied deleted vectors, default near-duplicate search missed a pair which exact search recovered. Forced ANN also changed recall score contributions. The precise CI curated-order inversion was not reproduced; it remains consistent with incomplete ANN membership, rather than proven as that incident's cause. These fixtures assert exact scoring/pair semantics, while ANN plans are covered separately. Fresh CI and the full baseline/candidate quality comparison remain merge gates.

Review recorded as COMMENT because this account owns the PR; no independent approval is claimed.

PR #502, opened 2026-09-07
perf(context): filter capture noise and scope content retrieval

cdeust:
Filter capture noise from methodology context and restrict fetch-by-ID access to the intended agent scope.

Includes current #501 slot/read fixes. Token ceilings remain unchanged; the separate 100-payload provider-token reduction claim is not established.

- Integrated current main (including #491) and every updated parent. Reviewed head: `af717ab76d936065233d7d4ec3202b5105ff00aa`.

The latest commit only triggers main-targeted security checks. Its tree `2abc1aa73492b4cbf2b1be51705aa73c302ed21a` is identical to `5000f75e5d1ee14c5dd62db709a99f86792dab2d`. The composed benchmark uses frozen commit `5000f75e5d1ee14c5dd62db709a99f86792dab2d`, whose tree matches final PR #502.


cdeust (commented):

cdeust (commented):
Reviewed the CI-trigger commit: git diff from 5000f75e5d1ee14c5dd62db709a99f86792dab2d is empty and Git tree 2abc1aa73492b4cbf2b1be51705aa73c302ed21a is unchanged. Previous source review and test evidence apply to this identical tree. Required main-targeted checks are now running; no protection bypass is used.

cdeust (commented):
Reviewed current head e7b86b481f9e29a68000e1cc5c46030a86047b5c including inherited test fix75fb7740.

The only source delta since the prior review is two PostgreSQL semantic test helpers. Both execute real production SQL with transaction-local exact scans; the scoring helper pins the real executor to the same pooled connection. All assertions are unchanged. Production runtime, scripts and benchmark trees are identical to the reviewed benchmark candidate.

Validation:35 native PostgreSQL tests passed (scoring-debias, near-duplicate pairs, indexed pools), zero skips; planner settings restore on both helpers. A rollback-only mutation removing the two real SQL auto-capture exclusions reverses curated/auto ranking and is rejected by the unchanged assertion. Ruff and formatting pass.

Empirical control: with2000 varied deleted vectors, default near-duplicate search missed a pair which exact search recovered. Forced ANN also changed recall score contributions. The precise CI curated-order inversion was not reproduced; it remains consistent with incomplete ANN membership, rather than proven as that incident's cause. These fixtures assert exact scoring/pair semantics, while ANN plans are covered separately. Fresh CI and the full baseline/candidate quality comparison remain merge gates.

Review recorded as COMMENT because this account owns the PR; no independent approval is claimed.

PR #512, opened 2026-09-08
fix(worktrees): create every worktree inside the repo under .claude/worktrees

cdeust:
## Why

`scripts/spawn-agent.sh` created its worktree as a sibling directory beside the repo, and `benchmarks/lib/bench_regression.sh` created its detached baseline worktree under `/tmp`. Worktrees outside the repo are invisible to Claude Code's own cleanup sweep and accumulate: on 2026-09-08, 43 detached and 33 branch worktrees under `/private/tmp/cortex-green-*` had to be removed by hand. Owner correction: the only sanctioned location is `<repo>/.claude/worktrees/<name>/`, already gitignored here via `.claude/*`.

## What

- `scripts/spawn-agent.sh`: worktree at `<repo>/.claude/worktrees/<agent>-<timestamp>`; warns if `.claude/worktrees/` is not gitignored; drops the now-unused `TARGET_NAME`.
- `benchmarks/lib/bench_regression.sh`: `mktemp -d` under `$REPO_ROOT/.claude/worktrees/bench-baseline-XXXXXX` instead of `${TMPDIR:-/tmp}`; the existing `worktree remove --force` cleanup is unchanged.
- `CLAUDE.md` "What NOT to do": the rule, with the measured incident.

## Verification

- `bash -n` on both scripts: ok.
- `python scripts/check_craftsmanship.py`: "Craftsmanship gate: OK".
- No Python touched; the doc-contract and lint jobs are the relevant CI checks.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -149,6 +149,14 @@ separate **cortex-viz** MCP (reads this same store read-only).
 - Do NOT write model caches under `/tmp` — the FlashRank incident (silently
   absent re-ranker, 6 benchmarks invalidated) came from exactly this.
   `HF_HOME`/`cache_dir` must be persistent.
+- Do NOT create git worktrees outside this repository — no `/private/tmp/cortex-*`,
+  no `../Cortex-wt-*` sibling, no `~/.claude/worktrees/`. The only location is
+  `.claude/worktrees/<name>/` (already gitignored via `.claude/*`): it is where
+  Claude Code's own worktree tooling (`isolation: worktree`, `EnterWorktree`,
+  `--worktree`) puts them and the only directory its cleanup sweep looks at.
+  Owner correction 2026-09-08, after 76 outside worktrees (43 detached + 33 on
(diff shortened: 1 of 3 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Fresh-context code-reviewer verdict on a13e845e. Verified live from the checkout: `.claude/worktrees` is ignored via `.gitignore:60` (`.claude/*`); a detached `git worktree add` nested inside the main worktree leaves `git status --porcelain` of the main checkout clean; the nested worktree carries its own `pyproject.toml`/`uv.lock`, so `uv run` from inside it resolves the baseline's project, not the parent's; the `if ! git check-ignore` block is safe under `set -euo pipefail`; `python scripts/check_craftsmanship.py` prints "Craftsmanship gate: OK". Non-blocking note: a future `git clean -fdx` from the main checkout would delete a live nested worktree's files (none exists in the repo today).

PR #534, opened 2026-09-09
fix(ast): capture explicit Python decorator calls

### LINKED ISSUE #372: feat(ast): decorator calls produce no graph edge — the decorated_definition arm in _walk_for_calls is an unfilled seam
## The gap

Calls made in a decorator are invisible to the code graph.

```python
@app.route("/api")
@retry(times=3)
def handler():
    work()
```

```
extract_calls_per_function -> {'handler': ['work']}
```

`route` and `retry` are call expressions in the source and produce no edge anywhere. For a graph whose purpose is answering "what calls what", decorator calls are not decoration: `@app.route(...)`, `@celery.task(...)`, `@pytest.fixture(...)` and `@click.command(...)` are how routes, tasks, fixtures and CLI entry points get registered. A consumer asking "what registers this route" gets nothing.

## Why this is filed rather than fixed in passing

Found while pinning the extractor walkers (#369). `_walk_for_calls` has a `decorated_definition` arm:

```python
elif ntype == "decorated_definition":
    stack.extend((c, scope) for c in reversed(child.children))
```

which is byte-identical to the `else` arm that follows it, so mutating it is undetectable and mutation reports its mutants as survivors. I initially removed it as dead code. That was wrong, and the maintainer caught it: the arm mirrors `_extract_python_children`'s dispatch, where the equivalent branch **is** load-bearing — that function handles exactly `function_definition`, `class_definition` and `decorated_definition` with **no catch-all**, so without it a decorated function disappears from definitions entirely.

The arm in the calls path is therefore a seam that was never filled in, not dead weight. It has been kept, and this issue is the missing behaviour.

## Design question to settle first

Where does a decorator call belong? Three defensible answers, and the choice is observable to every consumer:

1. **Attribute to the decorated function** — `handler` gains `route` and `retry`. Simple, and matches "this function's definition involves these calls". Conflates what the function calls at runtime with what runs at import time.
2. **Attribute to module scope** — needs a module-level pseudo-qname, which `extract_calls_per_function` has no concept of today.
3. **A distinct edge kind** — truest, most work, and requires the consumers (`ingest_codebase`, wiki reference pages) to understand a new relation.

Option 1 is the cheapest and probably right for a first pass, but it changes `extract_calls_per_function` output for every decorated function in every indexed repository, so it needs a deliberate decision rather than a drive-by.

## Acceptance criteria

1. A decision recorded for the question above (an ADR if option 3).
2. Decorator calls captured per that decision, with `@a.b(...)` resolving to the same basename convention `_callee_basename` already uses.
3. Tests: bare decorator (`@property`, no call), single call decorator, stacked decorators, decorator on a method inside a class, decorator on a class.
4. Mutation: the `decorated_definition` arm's mutants become killable, since the arm stops being behaviourally identical to the catch-all. Verify with the scoped runner over `mcp_server/core/ast_extractors.py`.
5. Downstream impact assessed by name: `ingest_codebase` and anything consuming its call edges, since existing graphs gain edges on re-index.

cdeust:
Explicit Python decorator calls now appear in the local call map for the decorated function, method, or class. Previously the decorated-definition branch behaved like generic traversal and lost these registration calls. Bare decorators retain their existing behavior. ADR-1058 records the deliberate choice to use lexical associations in the existing call-edge shape.

## Completion Ledger

| Issue acceptance | Evidence |
|---|---|
| Recorded ownership decision | Canonical `wiki/adr/cortex/1058-python-decorator-call-attribution.md` and generated mirror |
| Decorator calls and existing basename convention | Real parser tests cover called/stacked decorators, qualified callees and nested arguments |
| Bare/function/method/class cases | 257 focused AST, graph and mutation-registry tests passed |
| Decorated branch mutations killable | Scoped runner: 511 mutants, 488 killed, 23 pre-existing verified equivalents/unreachable; both decorator-dispatch mutants killed; obsolete seam exemptions removed |
| Named downstream assessment | Local parser/resolver improves; `ingest_codebase` and upstream wiki summaries consume the separate AP graph and are unchanged by this patch |

Measured real parser/resolver fixture: two existing body edges become five total edges, adding three registration edges while preserving both body edges. Full fixture, exact edge lists and source hash are committed in `docs/program/issue-372-decorator-call-coverage.json`. This is a static coverage measurement, not a runtime or retrieval-quality claim.

Full test collection (8,236 cases), pinned Ruff 0.16.6, Pyright 1.1.411 on the changed module, actionlint 1.7.12, uv 0.11.3 lock/export checks, docs/badge checks, craftsmanship and mirror integrity passed. Tests used a private Python environment and temporary backend state. Complete local evidence and limitations are in `docs/program/issue-372-decorator-calls.md`; required CI will verify the PR head.

Fixes #372.


cdeust (commented):
Codex-assisted independent review of head 3ae32f38402c713f44cd9fbf18d8240223eea723, posted through the author account as a COMMENT review. No blocking findings. Reviewed the explicit decorated-definition dispatch, existing basename and immediate-class conventions, stable deduplication and sibling/redefinition behavior. The canonical decision makes the lexical rather than runtime interpretation explicit, including the existing same-name class ambiguity. Verified the 2-to-5 real parser/resolver fixture preserves both body edges and adds three registration edges. Scoped mutation kills both former decorator-seam mutants; the 23 remaining cases are existing justified equivalents/unreachable cases with exact registry maintenance. Reviewed the named consumer boundary: AP-backed ingest_codebase and upstream wiki summaries do not change from this local extractor fix. Local tests, pinned static checks and committed evidence support the change. Ready for normal protected merge after required CI succeeds on this exact head.

Issue #538, opened 2026-09-09
scripts/setup.py installs an unpinned hand-written package list instead of the generated constraint file


cdeust:
## Problem

`scripts/setup.py` lines 200 to 212 install a hand-written list of loose version ranges:

```python
packages = [
    "mcp>=2.0.0",
    "pydantic>=2.0.0",
    ...
]
```

followed by a plain `pip install --target DEPS_DIR *packages`, with no pin, no hash and no reference to `uv.lock`.

This is the path taken by the SQLite backend on every OS and by the PostgreSQL backend on Windows, so it covers the default zero-config install, which is the most common one.

## Why it matters

`requirements/setup.txt` exists precisely to avoid this. Its header records the history: it "replaces a hand-written package list that had already drifted from pyproject.toml". `scripts/setup.sh` consumes it and is therefore reproducible; `scripts/setup.py` is not.

Concretely this path can resolve a set that `uv.lock` never validated, gets no hash verification, and will drift again exactly as the previous list did. It is also the path that is hardest to debug, since the resulting environment matches nothing CI ever tested.

No bug has been observed on it yet. This is about the guarantee, not a current failure.

## Suggested direction

Have `scripts/setup.py` install `requirements/setup.txt` with `--no-deps --require-hashes --upgrade`, the same way `scripts/setup.sh` now does, and delete the hand list. If a Windows-specific or SQLite-specific subset is genuinely needed, it should be its own generated constraint set in `scripts/pip_constraint_sets.py` rather than a list maintained by hand.

Issue #537, opened 2026-09-09
scripts/setup.sh step 5/7 reports [ok] after the model cache failed


cdeust:
## Problem

Step 5/7 of `scripts/setup.sh` prints a success label even when the pre-cache failed.

Observed verbatim during an install on 2026-09-09:

```
=== [5/7] Embedding model ===
  Pre-caching sentence-transformers model (one-time ~100MB download)...
  [\] running...  [|] running...  [/] running...Warning: model cache failed (
  ...
  Original error was: No module named 'numpy._core._multiarray_umath'
). Will download on first use.
[ok] Embedding model cached
```

The model was not cached. The `[ok]` line contradicts the warning printed immediately above it.

## Why it matters

Falling back to a lazy download on first use is a reasonable design, so the step should not be fatal. But the label must say what happened. As written, a reader scanning for `[ok]` lines concludes the cache is populated when it is not, and the real cause (here a stale numpy extension built for another interpreter) scrolls past unread.

## Suggested fix

Distinguish the two outcomes. Report `[ok] Embedding model cached` only when the pre-cache actually succeeded, and something like `[warn] Model not pre-cached, will download on first use` otherwise. The step stays non-fatal either way.

Issue #536, opened 2026-09-09
Gate: every hashed constraint install must pass --no-deps


cdeust:
## Problem

Nothing enforces the invariant that every `pip install --require-hashes -r requirements/*.txt` also passes `--no-deps`.

That absence is not theoretical. PR #332 fixed exactly this failure across CI, the Dockerfiles and `scripts/launcher_torch_cpu.py`, and recorded the mechanism in `docs/adr/ADR-0800-github-actions-test-suite-action-yml.md` lines 105 to 116. One call site was missed, `scripts/setup.sh` line 219, and it stayed broken until the plugin installer failed on it months later (fixed on branch `fix/setup-sh-no-deps`).

## Why it breaks

`pyproject.toml` lines 260 to 262 declare `[tool.uv] override-dependencies = ["mpmath>=1.4.1,<1.5"]`, which deliberately pushes mpmath past the `mpmath<1.4,>=1.1.0` bound `sympy 1.14.0` still declares in its metadata. uv's resolver honours the override; the `requirements.txt` format cannot carry it. Without `--no-deps`, pip re-derives the graph, sees only the raw conflict, and aborts with `ResolutionImpossible`.

The generated files are the complete, uv-resolved closure, so pip must never re-derive them. This is a property of every generated constraint file, not of one call site.

## Proposed check

A grep-level rule in the existing gates: any invocation combining `--require-hashes` with `-r requirements/*.txt` must also carry `--no-deps`. Scope it over `scripts/`, `.github/workflows/`, `.github/actions/`, the Dockerfiles and `.devcontainer/`.

Cheap to write, and it closes the exact hole that let the last fix be incomplete.

PR #535, opened 2026-09-09
test: close the provenance mutation gap (#389)

cdeust:
## Summary

Mutation testing on `mcp_server/core/provenance.py` left 43 genuine survivors out of 232 mutants (188 killed). That count was filed on 2026-08-08 and re-measured unchanged on 2026-09-09. This PR closes the gap: 29 survivors killed with new tests, and the remaining 13 registered as justified equivalents, each pinned to its exact removed/added pair as ADR-0771 requires.

Where the 29 came from:

- `_build_reason`, 13 mutants. No test asserted the wording it produces, because `grade_provenance`'s empty-outcome path hardcodes its own reason string instead of calling the helper. Its four branches are now pinned by direct unit tests carrying the boundary inputs, notably the three-ref join cap. One branch, `grade == UNVERIFIABLE` with `dead` empty, is unreachable through `grade_provenance`; the other three are reachable and are also pinned end-to-end by `TestGradeReportFields`.
- `grade_provenance`, 12 mutants. Every report field other than the grade (`memory_id`, `ref_counts`, `dead_refs`, `uncheckable_refs`, `reason`) went unread by the tests, and the documented least-favorable case, a ref absent from its verdicts dict, was never exercised.
- extraction, 3 mutants. The `continue` that skips a token could become a `break` unnoticed, the `rstrip` cut set could widen and eat a real path character, and a repeated artifact path was never deduped.
- `_ref_counts`, 1 mutant. The citation count was only ever asserted with a citation present, so the zero arm was free.

One production-code change, and it is a deletion: `dead_refs=[]` and `uncheckable_refs=[]` drop from the empty-outcome `ProvenanceReport`. They restate the dataclass defaults verbatim, and a pure-deletion mutant of a line like that carries no `added` text, which the equivalents registry rejects. Both values stay pinned by `TestGradeReportFields`.

Closes #389

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change existing behavior)
- [ ] Refactor (no functional change; rules/coding-standards.md compliance)
- [ ] Documentation only
- [x] Audit-finding closure (#389, provenance mutation coverage)

## Test plan

- `scripts/mutation_check.sh` on `mcp_server/core/provenance.py`: no unregistered survivors (29 newly killed, 13 registered equivalents).
- `pytest`: 8423 tests pass.
- `ruff check && ruff format --check`: clean.
[...]

PR #539, opened 2026-09-09
fix(setup): the installer's dependency step could neither install nor update

cdeust:
## Summary

`scripts/setup.sh` step 3/7 installs `requirements/setup.txt`, the hashed dependency closure exported from `uv.lock`. Its single `pip` invocation carried two independent defects, which together meant the plugin installer could not install its dependencies, and could not update them once it had. Both are fixed here.

No issue is closed by this PR; the failure was found while re-running the installer after PR #535. Three follow-ups are filed: #536, #537, #538.

### Defect 1, missing `--no-deps`

`pyproject.toml` lines 260 to 262 declare `[tool.uv] override-dependencies = ["mpmath>=1.4.1,<1.5"]` (mirrored at `uv.lock` line 30), which deliberately pushes mpmath past the `mpmath<1.4,>=1.1.0` bound `sympy 1.14.0` still declares in its own metadata. uv's resolver honours the override; the `requirements.txt` format cannot carry it. Without `--no-deps`, pip re-derives the graph, sees only the raw conflict, and aborts with `ResolutionImpossible`.

This is not a new discovery. PR #332 hit the same failure and fixed it everywhere else; the mechanism is written out in `docs/adr/ADR-0800-github-actions-test-suite-action-yml.md` lines 105 to 116. Every other consumer of a generated constraint file already passes `--no-deps`: `ci.yml` lines 185, 409, 527, 571, 783, 787 and 834, `release.yml` line 159, `scripts/launcher_torch_cpu.py` line 41, and the Dockerfiles. `scripts/setup.sh` line 219 was the one call site that fix missed.

Worth stating explicitly, since it was the first hypothesis: the installer does NOT bypass `uv.lock`. `requirements/setup.txt` is generated from it by `scripts/generate_pip_constraints.py`. Issue #253 is not involved.

### Defect 2, missing `--upgrade`

Surfaced only once the first was lifted. `pip install --target` skips every package whose directory already exists, prints `already exists. Specify --upgrade to force replacement` once per package, and the script then reports `[ok] Python packages installed`. Two consequences, both observed:

1. A re-run after the lock moves updates nothing, so a merge can never be reflected in a local install.
2. A re-run under a different interpreter keeps ABI-specific extension modules built for the old one. Here numpy's `_multiarray_umath` built for cpython-313 survived the switch to Python 3.14, so steps 5/7 and 6/7 failed with `No module named 'numpy._core._multiarray_umath'` and left sentence-transformers and the FlashRank reranker in `[FAIL]`.

The file is fully pinned and hashed, so the reinstall is deterministic and replacing is always correct here.

## Type of change

- [ ] Documentation only

## Test plan

This changes two flags in a shell script, so the verification is empirical rather than unit-level.

Defect 1, reproduced in isolation on a two-line extract of `requirements/setup.txt` holding only `mpmath==1.4.1` and `sympy==1.14.0` with their hashes:

- without `--no-deps`: `ResolutionImpossible`, reporting `sympy 1.14.0 depends on mpmath<1.4 and >=1.1.0`
- with `--no-deps`: `Would install mpmath-1.4.1 sympy-1.14.0`

Defect 2, verified end to end by running `scripts/install-plugin.sh` on macOS against the PostgreSQL backend:

- before: exit 1, about 140 `already exists` skip warnings, steps 5/7 and 6/7 red
- after: 7/7, exit 0, zero skip warnings, `deps/numpy/_core/` carrying the cpython-314 build, and `PYTHONPATH=deps python3 -c "import numpy"` succeeding under Python 3.14

Gates run locally: `bash -n`, `shellcheck -S warning`, `scripts/check_craftsmanship.py`, `scripts/check_doc_claims.py`, `ruff check`, `ruff format --check`. All green.

- [ ] New tests added for new behavior.
- [x] Mutation survival check: I considered what mutations would NOT be caught.
- [x] Manual verification of any UI / CLI / MCP-tool behavior changes.

On the unchecked box: no automated test guards either flag today, and adding one inline would be a grep over the repo rather than a unit test. That check is filed as #536, which is the durable fix for the class of defect rather than for this instance. Dropping either flag from this line would be caught by nothing in CI, which is precisely why the last fix could skip this site.

## Audit notes

Review passes are running; the verdict will be posted as a comment on this PR.

- Engineering review (code-reviewer / architect / refactorer / test-engineer): to follow
- Genius review (feynman / curie / popper / dijkstra / others): to follow
- Outstanding deferred findings: #536 (the missing guard for this invariant), #537 (step 5/7 reports `[ok]` after the model cache failed), #538 (`scripts/setup.py` still installs an unpinned hand-written list)

- [x] §2.2 Layer dependency direction preserved (no inward layer imports outward).
- [x] §3.2 No `any` (or `as any`, or untyped dicts at boundaries) in production code.
- [x] §4.1 No file > 500 lines (test files exempt).
- [x] §4.2 No function > 50 lines (dispatch tables exempt).
- [x] §7 Local reasoning preserved (no clever constructs that defeat single-function comprehension).
- [x] §8 Every numeric constant with 3 or more significant digits has a `source:` annotation.

## Breaking changes

None. The installer's observable contract is unchanged; it now honours it.

## Screenshots / logs

Step 3/7 before, truncated:

```
=== [3/7] Python dependencies ===
[ok] Python 3.14
Installing Python packages...
ERROR: Cannot install -r requirements/setup.txt (line 2569) and mpmath==1.4.1
because these package versions have conflicting dependencies.
ResolutionImpossible
[FAIL] Dependency install failed (see pip output above)
```

After both flags:

```
=== [3/7] Python dependencies ===
[ok] Python 3.14
Installing Python packages...
[ok] Python packages installed
...
=== [6/7] Verification ===
  [ok] PostgreSQL connection
  [ok] Extensions (pgvector, pg_trgm)
  [ok] PL/pgSQL recall_memories()
  [ok] sentence-transformers
  [ok] FlashRank reranker
```

## Reviewer checklist

- [x] CHANGELOG.md updated under the appropriate section.
- [ ] Documentation updated (README / SKILL.md / docs/).

On documentation: ADR-0800 already documents this mechanism, and the fix adds the rationale inline at the call site with `source:` markers pointing at ADR-0783 and ADR-0800. No doc page needed a change.


cdeust:
ZETETIC-REVIEW: APPROVE

Two passes ran on the original branch state. Both are recorded here in full, including what they refuted.

Engineering review: REQUEST-CHANGES on the first state.

- BLOCKER, upheld. `--upgrade` handed pip's delete-then-rewrite to `$DEPS_DIR`, the directory `scripts/launcher_deps_install.py` protects. That module resolves into a `.tmp-{pid}` scratch dir and commits entry by entry via `os.replace` with rollback, and its own docstring names the case: keeping numpy under a running MCP server out of the rmtree/replace path (ADR-0749). Verified independently: `.claude-plugin/plugin.json` lines 45 to 47 run the installer as the plugin postInstall, so a live session with the server holding `deps/` on `sys.path` is the normal invocation, not a corner case. Skip-if-present was safe under concurrency by construction. Fixed in f06db776 by removing the flag.
- SHOULD-FIX, upheld. The `source: ADR-0800` marker had drifted above a block it did not source. Moot now: the rationale left the code entirely, see below.

Falsification pass: APPROVE, with two of my claims refuted.

- Claim A survives. `ResolutionImpossible` reproduced from cached wheels, which excludes a stale index, on a pair that fails under any interpreter, which excludes Python 3.14, against a lock-derived file, which excludes issue #253. `uv.lock` `[manifest]` declares exactly one override, so mpmath is the only structural uv/pip divergence.
- Claim B survives, checked rather than trusted: `generate_pip_constraints.py --check` returns "requirements OK (13 checked)" and the embedded command is `uv export --locked`, which fails rather than re-resolving.
- Claim C survives. All twelve other generated-constraint consumers pass `--no-deps`; a repo-wide grep found no other executable counterexample.
- Claim D refuted as stated. `--upgrade` does purge a planted stale artifact, but "nothing else reintroduces a mismatch" is false: `launcher_deps_install.py` line 108 decides idempotence on dist-info version equality alone and ignores the ABI tag. `deps/` currently still holds four stale cpython-313 modules and `import websockets.speedups` fails under 3.14. So my earlier claim that the local install was realigned was wrong.
- Claim E refuted. `--upgrade` replaces package directories but not differently named top-level artifacts, and it would have masked the guard defect on every run. Filed as #540.

Net effect on the branch. `--no-deps` ships with its verification intact. `--upgrade` is out, for the concurrency hazard and because it would have hidden #540. The rationale moved out of the script into ADR-1059, since no decision has lived in the code since the Cortex refactoring; the wiki holds the decision and the code keeps a pointer.

Remaining diff against main in `scripts/setup.sh` is one flag and one `source:` line.

Deferred, all filed: #536 the missing gate for this invariant, #537 the false `[ok]` at step 5/7, #538 the unpinned list in `scripts/setup.py`, #540 the ABI-blind idempotence guard.

Issue #540, opened 2026-09-09
launcher_deps_install: the idempotence guard is ABI-blind, so stale extension modules survive an interpreter change


cdeust:
## Problem

`scripts/launcher_deps_install.py` line 108, `_entry_already_satisfied`, decides whether a resolved entry needs committing by comparing dist-info versions only:

```python
tmp_v = tmp_versions.get(key)
return tmp_v is not None and dest_versions.get(key) == tmp_v
```

The version can match while the compiled artifact does not. Extension modules carry an interpreter ABI tag in their filename, and the guard never looks at it, so after a Python upgrade every package whose version is unchanged keeps the build made for the old interpreter.

## Evidence

Observed in `deps/` on macOS after moving from Python 3.13 to 3.14, with four stale `cpython-313` extension modules still present: websockets, watchfiles, caio and cffi. `import websockets.speedups` fails under 3.14.

There is a second, related shape: `--target` installs also leave differently named top-level artifacts behind. `deps/_cffi_backend.cpython-313-darwin.so` sits beside the 314 build, so a move back to 3.13 would silently load the old one.

## Why the guard exists

The docstring is explicit and correct about its purpose, citing ADR-0749: it protects a locked, already-correct transitive dependency, numpy under a running MCP server being the named case, from ever entering the rmtree/replace path. That intent is right. The predicate implementing it is too weak.

## Why this matters now

PR #539 fixed the plugin installer's dependency step. The first attempt added `--upgrade` to `scripts/setup.sh`, which would have reinstalled everything on every run. That was dropped for two reasons: it bypasses the atomic-commit path this module exists to provide, and it would have repaired on every run whatever this guard reintroduces, hiding the defect rather than fixing it.

So the stale-deps problem is not fixed anywhere today. It belongs here.

## Suggested direction

Make the guard ABI-aware: treat an entry as satisfied only when the version matches AND the destination carries no extension module tagged for a different interpreter than the one running. Pruning orphaned top-level `*.so` artifacts that no dist-info claims would close the second shape.

PR #542, opened 2026-09-09
fix(setup): install requirements/setup.txt instead of a hand-written package list (#538)

### LINKED ISSUE #538: scripts/setup.py installs an unpinned hand-written package list instead of the generated constraint file

cdeust:
## Summary

scripts/setup.py's install_deps() built its own hand-written list of loose
version ranges ("mcp>=2.0.0", "pydantic>=2.0.0", ...) and passed it to
pip install --target with no pin, no hash, and no reference to uv.lock.
This is the path taken by the SQLite backend on every OS and by the
PostgreSQL backend on Windows: the default zero-config install.
requirements/setup.txt exists precisely to prevent this drift, and
scripts/setup.sh already consumes it (PR #539). install_deps() now
installs the same file with --no-deps --require-hashes and never
--upgrade, per ADR-1059, and the hand list is deleted.

Closes #538

## Type of change

## Test plan

- [x] All existing tests pass (pytest tests_py/ -q: 7988 passed, 290
      skipped on a full dev+sqlite+postgresql install; one unrelated
      failure, test_pg_driver_passes_in_dev_env, was an artifact of an
      earlier partial uv sync in my own venv, confirmed passing once
      --extra postgresql was added, not a regression from this diff).
- [x] New test added: tests_py/scripts/test_setup_py_installs_from_lock.py.
      It asserts the exact argv install_deps() hands to pip: source is
      "-r requirements/setup.txt" (not a Python list literal), --no-deps
      and --require-hashes are present, --upgrade is absent, and no
      hand-written name>=version specifier reaches pip. This test FAILS
      on origin/main (asserts "-r" in argv; main passes a list of loose
      specs instead) and PASSES after this change. Raw output of both
      runs is available on request.
- [x] Mutation survival check: a mutant that swaps --no-deps for --deps
      (removes the flag) or reintroduces --upgrade is caught by its own
      explicit assert ... in argv / assert ... not in argv line, so a
      partial mutation cannot slip past a single combined check.
- [x] Manual verification: scripts/generate_pip_constraints.py --check
      passes ("requirements OK (13 checked)"), confirming
      requirements/setup.txt is not stale relative to uv.lock.

## Audit notes

- Verified requirements/setup.txt (extras: postgresql, codebase,
  benchmarks) is a strict superset of the deleted hand list on both paths
  install_deps() runs (SQLite and PostgreSQL). mcp, pydantic,
  pydantic-settings, and numpy are pyproject.toml base dependencies
  (always exported regardless of extras); psycopg and pgvector come from
  postgresql; tree-sitter, tree-sitter-language-pack, and networkx from
  codebase; sentence-transformers, flashrank, and datasets from
  benchmarks.
- Checked whether a SQLite-specific subset (the [sqlite] extra,
  sqlite-vec) was needed instead, since install_deps() runs on the SQLite
  backend too. It is not: mcp_server/infrastructure/sqlite_store.py
  imports sqlite_vec as an optional dependency and degrades gracefully
  (vector search disabled) when absent, matching both the deleted hand
  list (which never included it either) and scripts/setup.sh (which has
  no SQLite-specific branch and installs the same requirements/setup.txt
  regardless of backend). No new generated constraint set was needed;
  reusing the existing setup.txt entry in scripts/pip_constraint_sets.py
  is correct.
- Read ADR-1059 (wiki/adr/cortex/1059-...md) for why --no-deps is
  required: pip cannot re-derive the graph [tool.uv]
  override-dependencies already resolved, and aborts with
  ResolutionImpossible on the mpmath/sympy bound. And for why --upgrade
  is forbidden: DEPS_DIR may be on a live MCP server's sys.path, and
  pip's delete-then-rewrite under --upgrade bypasses the atomic commit
  path of ADR-0749.

- [x] Section 2.2 layer dependency direction preserved. scripts/ sits
      outside the server to handlers to core/shared hierarchy; no import
      added.
- [x] Section 3.2 no any or untyped dicts introduced.
- [x] Section 4.1 no file over 500 lines (scripts/setup.py is 458 lines).
- [x] Section 4.2 no function over 50 lines (install_deps() is about 25
      lines).
- [x] Section 4.4 no function with more than 4 parameters.
- [x] Section 7 local reasoning preserved. No clever constructs added.
- [x] Section 8 source citations. Module docstring and inline comments
      cite ADR-1059 (flag rationale) and ADR-0749 (concurrency
      rationale).
- [x] Section 9 no dead code. The replaced hand list is deleted, not
      commented out.

## Breaking changes

None. Same install target directory. Same flag philosophy scripts/setup.sh
already ships (PR #539). Only the source of the package list and the pip
flags change.

## Reviewer checklist

- [x] No secret material or personal data in the diff.
- [x] ruff check, ruff format --check, check_craftsmanship.py,
      check_doc_claims.py, and generate_pip_constraints.py --check all
      pass on the final state.

Generated with Claude Code, https://claude.com/claude-code


cdeust:
ZETETIC-REVIEW: APPROVE

Reviewed the diff rather than the report.

What the change does, verified against `git diff origin/main...origin/fix/538-setup-py-from-lock`: the twelve hand-written version ranges leave `scripts/setup.py::install_deps`, replaced by `requirements/setup.txt`, the same generated hash-pinned closure `scripts/setup.sh` consumes. All three terms of the flag contract from ADR-1059 hold: `--no-deps` present, `--require-hashes` present, `--upgrade` absent. The ADR citation sits in the module docstring, which is the correct placement here, since a bare `# source: ADR-1059` above code trips ruff's ERA001 (the number parses as a subtraction, unlike one with a leading zero).

The gate is shown on both sides, which is what was asked. On origin/main the new test fails with `assert '-r' in [...]`, no `-r`, no `--no-deps`, no `--require-hashes`, a list of loose specifiers passed instead. After the change it passes. That is an external signal, not a self-assessment.

Two reservations, neither blocking.

The new test is hermetic: it mocks `run` and asserts the argv. That pins the flag contract, which is the thing most likely to be silently undone, but it does not prove a real install succeeds on the SQLite path or under Windows. The CI matrices answer that question and they are green here, so the gap is covered in aggregate rather than by this test.

The report set aside a failing `test_pg_driver_passes_in_dev_env` as an artefact of a partial `uv sync`. I did not take that at face value; CI's full matrix passing on this branch corroborates it independently.

CI: 24 pass, 4 skipping, zero failures.

Note for whoever rebases next: #543 and the pending #540 branch each add an entry under the same `## [Unreleased]` heading in CHANGELOG.md, so they will conflict there once this lands.

PR #543, opened 2026-09-09
fix(setup): stop labeling a failed model pre-cache [ok] (#537)

### LINKED ISSUE #537: scripts/setup.sh step 5/7 reports [ok] after the model cache failed

cdeust:
## Summary

Step 5/7 of `scripts/setup.sh` prints `[ok] Embedding model cached` even when the pre-cache failed. The pre-cache ran inside a python `try/except` that caught the exception, printed a warning, and fell off the end of the block — the subprocess always exited 0, so the shell's honest-looking `if spinner $!; then ok; else warn; fi` was fed a lie: the outcome was "reached the end of the step," not "succeeded at it."

Fix: remove the swallow so a failed import/load raises and the subprocess exits non-zero — the label is now derived from that exit code, never from control reaching the end of the step. The step's logic is extracted into `scripts/lib/precache_embedding_model.sh` (function-only, no top-level execution) so the outcome is drivable and assertable in isolation, since it wasn't reachable that way before. The step stays non-fatal by design (`|| true` in `setup.sh`, unchanged intent).

Closes #537

## Type of change

## Test plan

- `tests_py/scripts/test_precache_embedding_model_step.py` — two new tests, each sourcing the real `scripts/lib/precache_embedding_model.sh` with the production `ok()`/`warn()`/`spinner()` helpers and shadowing `sentence_transformers` on `PYTHONPATH` to force a real exception (the same failure mode from the issue transcript) or a real success, with no python3 mocking:
  - `test_forced_failure_prints_warn_not_ok` — asserts `[!!]` present, `[ok]` absent, exit code 1.
  - `test_forced_success_prints_ok_not_warn` — asserts `[ok]` present, `[!!]` absent, exit code 0.
  - Verified by hand against the pre-fix code path too: the same forced-failure harness against the old inline logic prints `[ok] Embedding model cached` — confirms the gate fails on `origin/main` and passes after this change.
- `bash -n scripts/setup.sh` — pass.
- `shellcheck -S warning scripts/setup.sh scripts/lib/precache_embedding_model.sh` — clean.
- `pytest tests_py/ -q` — full suite green (all tests passing, only expected sqlite/postgres-mode skips) after `uv sync --no-default-groups --extra dev --extra sqlite --extra postgresql`.
- `ruff check` / `ruff format --check` — clean.
- `python scripts/check_craftsmanship.py` — `Craftsmanship gate: OK`.
- `python scripts/check_doc_claims.py` — `doc claims OK`.
- [x] Mutation survival check: the two tests bracket both outcomes (exit 0 vs non-zero) of the sole branch point (`precache_embedding_model_step`'s `if spinner "$pid"`), so a mutant flipping that condition or reintroducing the swallow is caught by one of the two.
- [x] Manual verification: ran the fixed step end to end against the real `sentence-transformers` install (forced-success path) and against a shadowed `ModuleNotFoundError` (forced-failure path); raw output for both shown in the PR discussion / commit history.

## Audit notes

- Engineering review (self, engineer persona / Move 1-7 discipline applied): layer is `scripts/` (install tooling, not core/handlers/infrastructure) — no layer-boundary concern. Stakes classification: Low (install-tooling script, no auth/billing/schema/concurrency touch) but the fix itself required Move 3/4 discipline (root cause, not a symptom patch) given the explicit diagnosis.

- [x] §2.2 N/A — bash script, no layer imports.
- [x] §3.2 N/A — bash/python, no `any`.
- [x] §4.1 No file > 500 lines (new lib file is 44 lines; new test file is ~110 lines).
- [x] §4.2 No function > 50 lines (`precache_embedding_model_step` is ~25 lines).
- [x] §4.4 No function with > 4 parameters (2 params).
- [x] §7 Local reasoning preserved — no clever constructs; extraction makes the control flow more locally reasoned than before, not less.
- [x] §8 No new numeric constants.
- [x] §9 No dead code; no TODOs.

## Breaking changes

None. Step 5/7 output changes (an honest `[!!] Model not pre-cached, will download on first use` plus traceback replaces a false `[ok]`), the step remains non-fatal, and the overall install still proceeds identically either way.

## Reviewer checklist

- [x] CHANGELOG.md updated under `## [Unreleased]` / `### Fixed`.
- [x] Documentation updated — N/A (no README/SKILL.md claims about this step's exact output).
- [x] No secret keys or PII in the diff.
- [ ] CI passes on the latest commit (pending).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,19 @@ adheres to [Semantic Versioning](https://semver.org/).
   `requirements/setup.txt`, the same hashed closure `scripts/setup.sh`
   installs, with `--no-deps --require-hashes` and never `--upgrade`, for
   the same reasons recorded in ADR-1059. The hand list is deleted.
+- **`scripts/setup.sh` step 5/7 reported `[ok]` even when the embedding
+  model was never cached (issue #537).** The pre-cache ran inside a
+  `try/except` that printed a warning and then fell off the end of the
+  block, so the subprocess always exited 0 — the `[ok]` label was driven
+  by reaching the end of the step, not by the step succeeding. The
+  try/except is gone; a failed import or load now raises, so the label is
+  derived from the pre-cache subprocess's real exit code
+  (`scripts/lib/precache_embedding_model.sh`, extracted from `setup.sh` so
+  the two outcomes are drivable in isolation —
+  `tests_py/scripts/test_precache_embedding_model_step.py`). The step
+  still isn't fatal: a real failure now prints `[!!] Model not pre-cached,
+  will download on first use` plus the captured traceback, instead of
+  scrolling past unread under a false `[ok]`.
 
 - **The plugin installer could not install its dependencies (PR #539).**
   `scripts/setup.sh` step 3/7 installs `requirements/setup.txt`, the hashed
--- a/scripts/lib/precache_embedding_model.sh
+++ b/scripts/lib/precache_embedding_model.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Library: pre-cache the sentence-transformers embedding model and report
+# an honest [ok]/[!!] outcome derived from the pre-cache subprocess's exit
+# code — never from merely reaching the end of the step.
+#
+# Extracted from scripts/setup.sh step 5 (issue #537) so the two outcomes
+# can be driven and asserted in isolation: source this file, override
+# python3 on PATH, then call precache_embedding_model_step directly.
+# Function-only — sourcing this file runs nothing by itself.
+#
+# source: ADR-0783
+
+# precache_embedding_model_step <project_dir> <deps_dir>
+# Pre:  caller has already defined ok() and warn() (color-coded [ok]/[!!]
+#       line printers) and spinner() (animates a backgrounded pid, returns
+#       that pid's exit code via `wait`) — scripts/setup.sh defines all
+#       three before sourcing this file.
+# Post: prints exactly one [ok] or [!!] line, and on failure also prints
+#       the pre-cache subprocess's captured output so the real cause does
+#       not scroll past unread. Returns the pre-cache subprocess's exit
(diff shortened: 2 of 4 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Read the diff, not the report. The pre-cache ran inside a try/except that printed a warning and fell off the end of the block, so the subprocess always exited 0 and the [ok] label was driven by reaching the end of the step rather than by the step succeeding. The try/except is gone, the failure now raises, and the label derives from the real exit code. Non-fatal behaviour is preserved by an explicit `|| true` with its reason stated at the call site. Dropping the old `2>/dev/null` also surfaces the traceback, which is what made this morning's numpy cause scroll past unread.

The step was extracted to scripts/lib/precache_embedding_model.sh because it was not drivable in isolation; the test shadows sentence_transformers on PYTHONPATH to force a real ModuleNotFoundError rather than mocking python3.

Rebased onto b1d50d9e by me, keeping both CHANGELOG entries. CI: 24 pass, 4 skipping, zero failures.

PR #544, opened 2026-09-09
fix(launcher): make the deps idempotence guard ABI-aware (#540)

### LINKED ISSUE #540: launcher_deps_install: the idempotence guard is ABI-blind, so stale extension modules survive an interpreter change

cdeust:
## Summary

`scripts/launcher_deps_install.py::_entry_already_satisfied` (ADR-0749) decided whether a resolved `deps/` entry needed committing by comparing dist-info versions only. A version can match while the compiled artifact does not: extension modules carry an interpreter ABI tag in their filename (`*.cpython-313-darwin.so`), and the guard never looked at it, so after a Python upgrade every package whose version was unchanged kept the build made for the old interpreter. Reproduced on this machine after a 3.13→3.14 move: `deps/websockets/speedups.cpython-313-darwin.so` survived, `PYTHONPATH=deps python3 -c "import websockets.speedups"` failed under 3.14, and `deps/_cffi_backend.cpython-313-darwin.so` sat orphaned beside the freshly built `...cpython-314-darwin.so` (no owning dist-info, never revisited by the per-entry commit loop).

The guard is now ABI-aware: a version-matched entry is only treated as satisfied when it also carries no extension module tagged for a foreign interpreter ABI (walked recursively, so a whole package directory with a stale nested `.so` is force-committed through the existing atomic backup/replace path). A new sweep, `launcher_deps_fs.prune_foreign_abi_extensions`, removes orphaned top-level foreign-ABI files once a commit batch succeeds, closing the standalone-module shape. ADR-0749's original protection — never touch a locked, already-correct transitive dependency (numpy under a running MCP server) — is unchanged: the version check still runs first and short-circuits before any filesystem walk.

Decision recorded in ADR-1060 (`wiki/adr/cortex/1060-...md`, mirrored to `docs/adr/`).

Closes #540

## Type of change

- [ ] Documentation only

## Test plan

- [x] All existing tests pass: `pytest tests_py/` — 8447 passed, 8 skipped, 459 subtests passed.
- [x] New tests added for new behavior: `tests_py/scripts/test_launcher_deps.py` — direct unit coverage for `is_foreign_abi_extension`, `entry_has_foreign_abi_extension`, `prune_foreign_abi_extensions`, plus two end-to-end `_pip_install` tests (nested-package ABI mismatch forcing a real commit; orphaned top-level ABI file pruned after a successful batch). Both end-to-end tests were confirmed to FAIL on the pre-fix code (reverted the two source files, kept the tests, reran — both failed with `AttributeError`/`assert False is True`; restored the fix — both pass).
- [x] Mutation survival check: considered — the two `is None or ... != tmp_v` guard clauses and the `is_foreign_abi_extension`/`entry_has_foreign_abi_extension` boolean returns are each independently exercised by a dedicated true/false-case test; the untagged-suffix exclusion (`.abi3.so`/`.so`/`.pyd`) has its own explicit test so a mutant widening the regex to also flag those would be caught.
- [ ] Manual verification of any UI / CLI / MCP-tool behavior changes. N/A — no UI/CLI/MCP-tool surface changed.

## Audit notes

- Engineering review: self-verified via the craftsmanship gate (`scripts/check_craftsmanship.py` — OK) and `ruff check`/`ruff format --check` (both clean).
- Genius review: not run as a separate pass; the fix is a narrow, contract-driven addition to an existing, well-documented guard (ADR-0749) with direct reproduction evidence gathered on this machine before writing any code.

- [x] §2.2 Layer dependency direction preserved — both changed files are the pure-stdlib bootstrap layer described in their own module docstrings (`launcher_deps_fs.py`: filesystem primitives; `launcher_deps_install.py`: pip invocation + commit), no new imports crossing into `mcp_server`.
- [x] §3.2 No `any`/untyped dicts at boundaries — all new functions are fully typed.
- [x] §4.1 No file > 500 lines (`launcher_deps_fs.py` is 276 lines, `launcher_deps_install.py` is 218 lines).
- [x] §4.2 No function > 50 lines (largest new function, `prune_foreign_abi_extensions`, is ~12 lines of body).
- [x] §7 Local reasoning preserved — no reflection, no monkeypatching in production code, `current_suffix` is an explicit optional parameter (not a hidden global) so callers and tests can both reason about it directly.
- [x] §8 Every constant has a `# source:`/docstring `source:` annotation (the ABI-tag regex cites the `sysconfig.get_config_var` doc plus a measured verification on this machine).

## Breaking changes

None. The guard becomes strictly more conservative (fewer false "already satisfied" verdicts); the atomic commit/rollback path it routes through is unchanged (ADR-0749).

## Reviewer checklist

- [x] CHANGELOG.md updated under `[Unreleased]` / `### Fixed`.
- [x] Documentation updated — ADR-1060 added to `wiki/adr/cortex/`, mirrored to `docs/adr/` via `scripts/check_project_wiki.py --write`, `wiki/manifest.json` updated.
- [x] No secrets, tokens, or personal data in the diff.
- [ ] CI passes on the latest commit — pending, will confirm once CI reports.


cdeust:
ZETETIC-REVIEW: APPROVE

Read the diff. The predicate is derived from the running interpreter's own EXT_SUFFIX via sysconfig, with an importlib.machinery fallback for builds where that var is unset, and it deliberately excludes the untagged .abi3.so/.so/.pyd forms, which are stable-ABI or source-shared and never foreign. That exclusion is the part most likely to have been got wrong, and it is right.

The ADR-0749 intent is preserved rather than weakened: the version comparison still runs first and cheap, and the filesystem walk only follows it, so the guard still keeps an already-correct transitive dependency out of the rmtree/replace path. Orphan pruning is best-effort, top-level only, and runs after a whole batch commits, in the same place prune_superseded_dist_info already runs.

The gate is external and shown both ways: the two tests fail on unmodified origin/main with AttributeError (no current_extension_abi_suffix) and pass after. The defect itself was reproduced on real files before any code was written, not inferred.

ADR renumbered 1060 to 1061 after ADR-1060 was taken on main mid-flight, with check_project_wiki.py confirming the mirror agrees.

Accepted risk, stated in the ADR: the tag regex is CPython-specific, so a non-CPython interpreter sees nothing to classify and the predicate degrades to a no-op rather than a false positive.

CI: 24 pass, 4 skipping, zero failures. Merges cleanly onto f9addd77.

PR #545, opened 2026-09-09
fix(hooks): close three decision-gate review blockers

cdeust:
## Summary

PR #541 (feat/decision-gate-hook, merged as 74aeee44) shipped `mcp_server/hooks/decision_gate.py` with three review blockers, reproduced and fixed here.

1. Crash on unreadable input: `Path.read_text` caught only `OSError`; a `UnicodeDecodeError` (a `ValueError`) or a null byte in the path crashed the hook and blocked every edit to that file. Both read call sites now share a `_read_text` helper catching `(OSError, ValueError)`.
2. No lexical awareness: a module docstring followed by a licence block, and a heredoc body containing commented example data, were both blocked as if they were decisions. `mcp_server/hooks/_decision_gate_lex.py` (new) scans Python comments via `tokenize.COMMENT`, tolerates a leading docstring as header material, and skips shell heredoc bodies.
3. Grandfathering broke on reflow: keying a run by its exact joined text meant inserting or removing a bare marker line inside an existing block reset the key and lost the grandfather. It now compares the set of existing body-comment line texts against the candidate run.

Two further findings from the review, resolved:
- `is_test_path` now recognizes `tests_js` and `*.test.*`/`*.spec.*` filenames uniformly across languages (a Python test file was exempt while the same shape under a JS test root was not).
- `COMMENT_MARKERS` narrowed to the `#`-comment languages the eight-line threshold was actually measured against on the tracked tree (`.py`: 1418 files, `.sh`: 18 files); `//` and `/* */` markers had zero non-test tracked instances and are dropped until measured.

The case the gate exists for still blocks: a 10- or 12-line rationale block added to a script (regression test added).

## Why a new PR

#541 merged to `main` (74aeee44) before this fix was ready; the remote branch was deleted on merge. This is a follow-up fix branched from current `main`, not an amendment to the closed PR.

## Test plan

- `tests_py/hooks/test_decision_gate.py`: 19 tests, one per blocker reproduced against `evaluate` pre-fix and passing post-fix, plus two through the real entry point (`test_entrypoint_exits_two_on_a_blocked_write`, `test_entrypoint_never_crashes_on_an_unreadable_file`) so the exit code is covered. `test_still_blocks_a_ten_to_twelve_line_rationale` pins the non-negotiable case.
- `.venv/bin/python -m pytest tests_py/ -q`: 8005 passed, 290 skipped, 1 failed (`test_pg_driver_passes_in_dev_env`, pre-existing, environment-dependent, unrelated to this change — this dev venv was installed with `--extra sqlite` only, not `--extra postgresql`).
- `ruff check` / `ruff format --check`: clean.
- `scripts/check_craftsmanship.py`: OK (`decision_gate.py` 247 lines, new `_decision_gate_lex.py` 83 lines).
- `scripts/check_doc_claims.py`: OK.
- `scripts/check_project_wiki.py`: OK, `docs/adr/` mirror regenerated via `--write`.

`wiki/adr/cortex/1060-...md` carries a revision section describing what changed and why; `CHANGELOG.md` Added/Fixed entries describe the shipped behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,19 +20,45 @@ adheres to [Semantic Versioning](https://semver.org/).
   code file. A `source:` line never counts toward a run, and the refusal
   names `wiki_adr` and the pointer form to leave behind. Exempt: the file
   header, defined as a run with nothing executable before it rather than by
-  a line number; test files; and any block already in the file, so a legacy
-  file stays editable elsewhere. `CORTEX_DECISION_GATE=off` overrides one
-  call. See ADR-1060.
+  a line number; test files (now including `tests_js` and `*.test.*`/
+  `*.spec.*` filenames, not just the Python shapes); and any block already
+  in the file, compared by its set of comment-line texts rather than exact
+  position, so a rewrap of an existing block is never mistaken for new
+  prose. Enforced only for the `#`-comment languages the eight-line
+  threshold was actually measured against on the tracked tree — `.py`,
+  `.sh`, `.bash`, `.zsh`, `.rb` (1418 and 18 tracked files respectively for
+  the two with volume); `//` and `/* */` languages have no tracked
+  instances yet and are not enforced. `CORTEX_DECISION_GATE=off` overrides
+  one call. See ADR-1060.
 
 ### Fixed
 
 - **`scripts/setup.py` installed an unpinned, hand-written package list
   instead of the generated constraint file (#538).** This is the path taken
   by the SQLite backend on every OS and by the PostgreSQL backend on
-  Windows — the default zero-config install. `install_deps()` now resolves
+  Windows, the default zero-config install. `install_deps()` now resolves
   `requirements/setup.txt`, the same hashed closure `scripts/setup.sh`
   installs, with `--no-deps --require-hashes` and never `--upgrade`, for
   the same reasons recorded in ADR-1059. The hand list is deleted.
+
+- **The decision-gate hook crashed on unreadable input and misread strings,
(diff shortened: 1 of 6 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

I verified the behaviour myself on the branch rather than reading the report, driving `evaluate` on the five cases that mattered:

- a twelve-line rationale block added to a shell script is still refused (exit 2). This was the non-negotiable constraint, since a fix that let it pass would have deleted the feature rather than repaired it;
- a binary file under a .py suffix returns 0 without raising;
- a module docstring followed by an eight-line licence block is allowed;
- a heredoc body containing commented example lines is allowed;
- a pure reflow of a block already present in the file is allowed.

All five were reproduced as defective before, on the version that is currently on main.

One further defect was found by CI's type gate after the agent's report, not by the review or the tests: the lexer's fail-open handler named `tokenize.TokenizeError`, which does not exist. An except clause is evaluated only when it fires, so the handler meant to keep the gate open on unparsable Python raised AttributeError at exactly the moment it was needed. Nineteen tests passed over that line without reaching it. Fixed in 27bdbbd9, with two tests that drive the path (an unterminated bracket makes tokenize raise) and fail with AttributeError against the misspelled name.

The lexical work was extracted to _decision_gate_lex.py rather than inflating decision_gate.py past the 300-line cap.

CI after re-running an infrastructure-cancelled SQLite job: 24 pass, 4 skipping, zero failures. Merges cleanly onto eae0978d.

PR #551, opened 2026-09-09
docs: reflect the v4.20.0 LongMemEval-S measurement in the README, with its artifacts

cdeust:
## Summary

The README's retrieval table cited only the v4.14.1 pre-tag run. This adds the v4.20.0 figure beside it, backed by artifacts produced for this change on the exact tagged tree, and scopes the Verification lead-in so it no longer claims an ablation campaign behind a figure that has none.

### The measurement

`benchmarks/reproduce.sh --only longmemeval --no-ablation`, run from a detached worktree at the v4.20.0 commit (86251ab8), isolated ephemeral `pgvector/pgvector:pg16` container, reranker loaded, consolidation disabled, n=500.

| | v4.14.1 | v4.20.0 |
|---|---|---|
| Recall@10 | 98.2% | 97.8% |
| MRR | 0.9167 | 0.905 (0.9047) |

Artifacts committed under `benchmarks/results/repro/20260909-v4.20.0-longmemeval-s/`: `longmemeval-s.json` (overall and per-category metrics, elapsed 1903.6 s, zero consolidation calls, embedded manifest), `MANIFEST.json` (git SHA, dataset sha256, image, embedding model revision, reranker state, Python 3.12.11), `START_SNAPSHOT.json`. The standalone MANIFEST.json has no dirty flag field, but the JSON's embedded `manifest.repro.git_dirty` is `false`, and the README now states dirty=false from it; the first cut of this body claimed no flag existed, which the fresh-context review corrected.

The run's own floor check: Recall@10 0.978 against the July floor 0.982 is within the 0.005 tolerance and passes; MRR 0.9047 against 0.914 is 0.0093 below and fails, which the script treats as non-blocking by design. `docs/agent-guidance.md` already records that `main` no longer clears those floors and that the release gate is `--no-regression` against `origin/main`. LoCoMo and BEAM have no v4.20.0 figure; both legs of the earlier full run were killed by the host for memory pressure, and this change measures only the leg it cites.

### README changes

- The retrieval table gains a v4.20.0 column with its own provenance paragraph: artifact JSON, manifest, code SHA, run conditions, delta against v4.14.1, and the floor-check outcome stated as the script reports it.
- The Verification lead-in now reads that the v4.14.1 figures are backed by the ablation campaign and that the v4.20.0 figures are a single measurement without one. The previous sentence covered "every benchmark headline above", which would have been false once this column existed.

### Why this run and not the earlier one

A full `reproduce.sh` run on the same tree earlier today completed this leg with the same numbers (0.905 / 0.978) before the host killed it, and its `longmemeval-s.json` was then lost. The surviving run log could not be committed: `.gitignore` excludes `benchmarks/results/**/*.log`, and citing an artifact absent from the tree is not allowed here. Re-measuring was the only honest route, and the first attempt at that exposed a script defect fixed separately in #550 (`--only longmemeval-s` silently selected nothing).

## Type of change

- [x] Documentation
- [x] Benchmark artifact

## Test plan

- Every relative link resolves, including the three new artifact paths.
- `scripts/check_doc_claims.py`, `scripts/check_version_surfaces.py` (17 sites): green.
- `tests_py/scripts/test_codex_plugin_contract.py` and `test_version_surfaces.py`: 33 pass.
- Table cells were asserted against the JSON in the script that edited the README (0.978 exact; MRR rounds to 0.905).

- [x] Manual verification: link resolution and cell-to-artifact assertion.

## Audit notes

Self-review plus the artifacts. A fresh-context review is requested before merge.

## Reviewer checklist

- [x] CHANGELOG.md unchanged; documentation and artifact addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ v4.14.1, 2026-07-14: [artifact JSON](benchmarks/results/repro/20260714-v4.14.1-p
 This is the run the ablation campaign in [Verification](#verification) was built around.
 
 v4.20.0, 2026-09-09: [artifact JSON](benchmarks/results/repro/20260909-v4.20.0-longmemeval-s/longmemeval-s.json) and its
-[manifest](benchmarks/results/repro/20260909-v4.20.0-longmemeval-s/MANIFEST.json); [code SHA](https://github.com/cdeust/Cortex/commit/86251ab8fc27a18f80f9b09b99a75f3b60edd9cb).
-A single run of the LongMemEval-S leg alone (`benchmarks/reproduce.sh --only longmemeval
+[manifest](benchmarks/results/repro/20260909-v4.20.0-longmemeval-s/MANIFEST.json); [code SHA](https://github.com/cdeust/Cortex/commit/86251ab8fc27a18f80f9b09b99a75f3b60edd9cb),
+dirty=false. A single run of the LongMemEval-S leg alone (`benchmarks/reproduce.sh --only longmemeval
 --no-ablation`) in an isolated ephemeral PostgreSQL container, reranker loaded, consolidation
 disabled. Against v4.14.1 the change is 0.4 points of Recall@10 and 0.012 of MRR. The run's own
 floor check reports Recall@10 within the 0.005 tolerance of the July floor (0.982) and MRR
--- a/README.md
+++ b/README.md
@@ -188,14 +188,28 @@ from it.
 
 **LongMemEval**: 500 human-curated questions buried in about 40 sessions of history.
 
-| Historical Cortex run (v4.14.1) | Result |
-|---|---|
-| Recall@10 | **98.2%** |
-| MRR | **0.9167** |
+| | v4.14.1 (historical) | v4.20.0 (current release) |
+|---|---|---|
+| Recall@10 | **98.2%** | **97.8%** |
+| MRR | **0.9167** | **0.905** |
 
-Historical single run from v4.14.1, not a v4.20.0 performance result: n=500, clean database,
-consolidation disabled. [Artifact JSON](benchmarks/results/repro/20260714-v4.14.1-pretag/longmemeval-s.json);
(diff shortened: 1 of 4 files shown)


cdeust:
ZETETIC-REVIEW: APPROVE

Verdict on head e7d5bcf9.

An independent fresh-context reviewer read every number in the README paragraph against the committed artifacts, field by field: overall_recall10 0.978 exact; overall_mrr 0.9046547619047619 rounding to 0.905; the deltas against the committed v4.14.1 artifact, 0.4 points of Recall@10 and 0.01205 of MRR; consolidation_call_count 0; elapsed_s 1903.6. MANIFEST.json's git_sha equals the commit the README links and equals `git rev-parse 'v4.20.0^{commit}'`; reranker_state loaded and reranker_active true; pg_image pgvector/pgvector:pg16. The floor wording was checked against reproduce.sh's constants (FLOOR_LME_R10 0.982, FLOOR_LME_MRR 0.914, tolerance 0.005, delta >= -tol) and against docs/agent-guidance.md lines 47 to 51: Recall@10 at -0.004 passes, MRR at -0.0093 fails, non-blocking by the script's own comment. No v4.20.0 ablation artifact exists, so the scoped Verification lead-in holds. Every relative link resolves; check_doc_claims.py and check_version_surfaces.py pass; the contract and version-surface tests pass 33 of 33. The diff touches README.md and the three artifact files only.

The reviewer's one finding, a NIT, is closed in e7d5bcf9: the standalone MANIFEST.json has no dirty flag, but the JSON's embedded manifest.repro.git_dirty is false, so the README now states dirty=false from it and the PR body no longer claims the flag is absent. I verified that field myself before editing.

The re-measurement reproduces the earlier killed run's figures to the third decimal (0.905 / 0.978), and matches the 2026-09-08 comparison on main, so v4.20.0 shows no regression against main under the committed 0.005 tolerance; it sits below the July absolute floors, which docs/agent-guidance.md already records.

CI on this head: 24 pass, 4 skipping, zero failures.

PR #569, opened 2026-09-15
fix(core): remove os/pathlib from wiki_coverage cluster (#560, 3/3)

cdeust:
## Summary

Third and final PR of the multi-PR #560 fix: the wiki filesystem cluster (6 of the 17 `mcp_server/core/` modules that imported os/pathlib). Together with #564 (2 modules) and #568 (9 modules) together they complete #560.

- `wiki_groomer.py`, `wiki_file_doc_skeleton.py`: pathlib/os.path usage was pure string manipulation (no I/O) - replaced with local helpers pinned equal to `PurePosixPath`/`os.path.splitext` by new tests.
- `wiki_drift.py` + new `mcp_server/infrastructure/wiki_drift_fs.py`: real filesystem I/O (read a page, mtime, walk the wiki tree, walk a source tree for file-exists checks) moved to infrastructure; core goes through `configure_wiki_drift_filesystem(...)`. Public function signatures are unchanged, so existing handler call sites and all 11 existing tests needed zero changes.
- `wiki_coverage_dashboard.py` + new `mcp_server/infrastructure/wiki_dashboard_fs.py`: same treatment for gap counting, kind-page counts, domain-dir listing, and the actual dashboard file writes.
- `wiki_axis_registry.py` + new `mcp_server/infrastructure/wiki_axis_fs.py`: the `_schema/` file reads moved to infrastructure via `configure_schema_file_reader(...)`.
- `wiki_coverage.py` (1330 lines, the largest module) + new `mcp_server/infrastructure/wiki_coverage_fs.py`: 5 functions doing real I/O (stat, listdir, walk, open+read) moved behind one `configure_wiki_coverage_filesystem(...)` seam. `_SKIP_DIRECTORIES` now imports the shared frozenset already extracted for `wiki_drift_fs.py` instead of duplicating it (infrastructure may not import core, so it lives in `mcp_server/shared/wiki_skip_directories.py`).

**Seam-fragility audit** (per the CI failure caused by PR #568's reranker seam being called from a bare CI script that bypassed every composition root): grepped `scripts/`, `.github/`, `benchmarks/`, and `mcp_server/hooks/` for every function and `configure_*` name this PR introduces - zero hits outside `mcp_server/__main__.py`, `tests_py/_composition_root_wiring.py`, and handlers. Unlike the reranker seam, none of these four wiki seams have bare-script exposure today, so their fail-fast "raise if unconfigured" design is safe (a wiki audit silently returning "no drift found" on misconfiguration would be a worse failure than a loud crash). Also re-audited PR #564's seams: both already safe (graceful degrade or no I/O at all).

## Test plan

- `python scripts/check_craftsmanship.py --base origin/main` returns OK
- grep for os/pathlib in these 6 files returns empty
- `ruff check .` and `ruff format --check .` are clean
- `pyright mcp_server/` reports 0 errors
- `python scripts/verify_mcp_hosts.py -- hypermnesia-mcp` passes on all 6 client/profile combinations
- Full `pytest -q`: 8447 passed, 165 skipped, 0 failed

Refs #560 (the last of the three PRs to merge, #568, closes the issue)


cdeust:

Reviewed at 88b7a9ce, rebased on 427b923a.

1. **Production regression in the background consolidation.** The six wiki filesystem seams (`configure_wiki_coverage_filesystem`, `configure_wiki_drift_filesystem`, `configure_dashboard_filesystem`, `configure_default_wiki_root`, `configure_schema_file_reader`) are wired only in `mcp_server/__main__.py` and `tests_py/_composition_root_wiring.py`, and they raise when unwired. `mcp_server/hooks/consolidate_background.py`, spawned by SessionStart, imports `handlers.consolidate` directly and runs it with `"wiki": True` (line 79). That path goes through `handlers/consolidation/wiki_maintenance.py`, which reaches `core/wiki_coverage` and `core/wiki_coverage_dashboard`, so on this branch every autonomous consolidation fails in the hook process. `scripts/regenerate_file_docs.py` reaches `core/wiki_coverage` the same way. CI cannot see this because the suite always wires the seams in conftest.
2. **The design multiplies wiring sites.** Together with #568, every new seam must be added by hand to `__main__`, 14 hook entry points, the benchmark bootstrap and the scripts. One composition-root function should wire every seam (environment reader, reranker filesystem, wiki filesystem), with every entry point calling it. A test must enumerate the entry points mechanically (hook commands in `.claude-plugin/plugin.json`, `benchmarks/**/run_*.py`, `scripts/*.py` reaching core, `__main__`) and assert that each one calls it. A subprocess test must run `consolidate_background` with `wiki: True` on a throwaway store without the test conftest, and assert that it completes.

cdeust:
ZETETIC-REVIEW: APPROVE

Re-reviewed at f77040ad, base 427b923a. Both points of the earlier REQUEST_CHANGES are addressed.

- The six wiki seams still raise when unwired. They are now wired by a single idempotent `wire_composition_root()` in `mcp_server/hooks/wiring.py`. Every entry point calls it: `__main__`, `scripts/launcher.py`, every hook's main block, every script and benchmark runner that reaches core, and the test session. It lives in hooks/ because that is the only layer that may import both core and infrastructure and also be imported by hook processes. The CI craftsmanship gate refused the earlier top-level placement.
- Measured on the hook path: `python -m mcp_server.hooks.consolidate_background` on a throwaway SQLite store logs `dashboards=written=0`. The same hook body run without the wiring logs `dashboards=error: RuntimeError: ... not configured`. The hook's summary line no longer defaults the dashboard status to `ok`.
- `test_composition_root_wiring_coverage.py` enumerates entry points mechanically and requires an actual call, not a mention. The subprocess test has no wall-clock verdict.
- `post_tool_capture` wires after the capture-mode gate, so excluded events still import nothing from core or infrastructure (W3-1b).
- Cost of the wiring import per hook process: 24-35 ms cumulative (`-X importtime`, 3 runs).
- Gates on the final state: full suite 8617 passed (one coverage-test false positive fixed, the hook tests re-run 472 passed). pyright 0 diagnostics. The craftsmanship gate passes with `--base origin/main`, as CI runs it. CI is green on all 24 required checks.

PR #581, opened 2026-09-16
feat(wiki): expose the Path B draft tools over MCP (#579)

cdeust:
## Summary

`wiki_get_draft` and `wiki_refine_draft` (Path B of the wiki pipeline, ADR-0467) existed as handlers with no registration: `mcp_server/tool_registry_wiki.py` registered ten other wiki tools and no commit in the repository's history ever registered these two, so no client could call them and `wiki_get_draft`'s own instructions told the caller to use a tool `tools/list` never advertised.

They now register from `mcp_server/tool_registry_wiki_drafts.py`, wired into `__main__`'s `merged_schemas()` and `register_all()`. The pair lives in its own module because `tool_registry_wiki.py` stands at 266 lines against the 300-line cap and two registrations in that file's style cost about sixty (ADR-1066). Reading a draft is `READ_ONLY`; submitting refined prose is `NON_IDEMPOTENT_WRITE`, since each call inserts an audit memo; both are `interactive` in `handlers/latency_class.py`. Neither joins `LEAN_TOOL_NAMES`, so the lean profile is unchanged.

The standalone count moves from 52 to 54, and from 55 to 57 with both upstream integrations: the pinned `tests_py/test_main.py::test_standalone_baseline_is_54_tools`, `docs/mcp-tools.md`, the seven other documents `scripts/check_doc_claims.py` scans, and the `docker_smoke.sh` floor.

Closes #579

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test plan

- New `tests_py/handlers/test_wiki_draft_tools.py`: both names on `list_tools()` with their annotations; the listing arguments (`list_pending`, `kind`, `limit`) and the content arguments (`lead`, `sections`, `rationale`, the `synth_model` default) reaching the handlers through `mcp.call_tool`; and one `wiki_refine_draft` call landing in a SQLite store.
- `tests_py/handlers/test_tool_schema_parity.py` already requires every registered wrapper's parameters to equal its handler's `inputSchema` properties exactly, and now covers these two.
- Full suite: 8658 passed, 8 skipped. The 5 errors are `tests_py/scripts/test_installer_*` (added by #575), which shell out to `python -m pip` in the venv; this machine's uv venv ships no pip (`No module named pip`). They touch none of the files in this diff.
[...]

Issue #579, opened 2026-09-15
[bug] wiki_get_draft and wiki_refine_draft (Path B) are defined but never registered as MCP tools


cdeust:
### What happened

`mcp_server/handlers/wiki_refine.py` defines Path B of the wiki pipeline (ADR-0467): `handler_get` (`wiki_get_draft`) hands an LLM a draft, its source claims and the kind contract, and `handler_refine` (`wiki_refine_draft`) takes the refined prose back. Neither is registered as an MCP tool. `mcp_server/tool_registry_wiki.py` registers ten wiki tools, none of them these two, and `git log -S wiki_refine` over the `tool_registry*.py` files finds no commit that ever registered them.

The only caller in the tree is `tests_py/handlers/test_wiki_pipeline_sqlite.py::test_refine_lists_pending_drafts` (for `handler_get`) and, since #578, `tests_py/handlers/test_wiki_refine_confidence.py` (for `handler_refine`). The maintainer's PostgreSQL store holds one `refined_llm` memo, written on 2026-04-14 by a `test_runner_v1` run; all 546 drafts carry `synth_model` `template_v1`.

`handler_get`'s instructions tell the caller to "call wiki_refine_draft", a tool no client can see.

### Expected behavior

One of the two, decided by the owner: register both tools (with their `tool_profiles.py` and `latency_class.py` entries and the `docs/mcp-tools.md` count), or remove the module and its tests with an ADR that supersedes ADR-0467.

### Evidence required to close

Either the two tools listed by the server's `tools/list`, or the module gone with no remaining reference outside `wiki/adr` and `docs/adr`.

Issue #578, opened 2026-09-15
[bug] wiki_refine_draft sets every refined draft's confidence to 0.85, lifting it past the curator's approve threshold


cdeust:
### What happened

`wiki_refine_draft` (`handler_refine`, `mcp_server/handlers/wiki_refine.py:300`) overwrites the draft's confidence with a fixed `0.85` every time an LLM submits refined prose, and records the audit memo with the same `0.85` (line 315). No source backs the value: ADR-0467 preserves the file as it was during #514 and records no threshold choice.

The template synthesizer sets a draft's confidence to the mean confidence of its source claims, or `0.4` when no claim carries one (`mcp_server/core/draft_synthesizer.py:271-277`). The curator uses that number as a gate: below `MIN_CONFIDENCE_APPROVE = 0.6` a draft gets a reason and can at best be held, below `MIN_CONFIDENCE_HOLD = 0.4` it is rejected (`mcp_server/core/draft_curator.py:15-16, 99-110, 138-141`). `wiki_curate` then moves approved drafts toward `wiki_compile`.

So rewriting a draft's prose lifts it above the approve threshold whatever the evidence behind its claims. A draft synthesized at 0.4 from weak claims clears the confidence check of `evaluate_draft` once refined, although no claim changed.

### Reproduction steps

1. Run `wiki_extract` then `wiki_synthesize` on any store; pick a pending draft and note its `confidence`.
2. Call `wiki_refine_draft` with that `draft_id` and a new `lead`.
3. Read the draft again: `confidence` is `0.85`, and `evaluate_draft` no longer reports "below approve threshold" for it.

### Expected behavior

Refinement changes the wording of a draft and leaves its confidence as the source claims set it. The audit memo records the draft's confidence as it stands, not a constant.

### Evidence required to close

A test on the SQLite backend that refines a draft whose confidence is below `MIN_CONFIDENCE_APPROVE` and asserts that its confidence is unchanged, that the `refined_llm` memo carries that confidence, and that `evaluate_draft` still reports the approve-threshold reason.

cdeust:
Correction on the reach of this defect, checked after opening the issue.

[...]

PR #584, opened 2026-09-16
fix(launcher,wiki): install without a pip module, refuse an empty refine (#582, #583)

### LINKED ISSUE #582: [bug] the launcher cannot install anything from an interpreter without pip
### What happened

The launcher installs every dependency with `sys.executable -m pip` (`scripts/launcher_pip.py::_pip_command`). When the interpreter running the launcher has no `pip` module, every install fails at once:

```
[cortex-launcher] dependency install failed:
  .../.venv/bin/python: No module named pip
```

A virtual environment created by `uv venv` (and `python -m venv --without-pip`) has no pip by design, so a user who launches Cortex from such an interpreter cannot install the plugin's private `deps/` at all. The failure is reported through the normal path and no success stamp is written, so nothing is corrupted, but nothing installs either.

Measured on 2026-09-16 on the maintainer's machine, Python 3.13.7 in a `uv venv`: the 11 tests of `tests_py/scripts/test_installer_rerun_keeps_entries.py` (6) and `tests_py/scripts/test_installer_upgrade_single_version.py` (5) error in setup with that message, because they call `scripts/launcher_deps.py` with `sys.executable`. Under a Homebrew Python 3.14.4, which has pip, the same tests pass.

### Expected behavior

The launcher installs with the pip that ships with the standard library when the interpreter has no pip module. CPython bundles a pip wheel for `ensurepip` (`ensurepip/_bundled/pip-*.whl`), and that wheel can be run in place without installing anything:

```
$ .venv/bin/python -c "import ensurepip, pathlib; print(pathlib.Path(ensurepip.__file__).parent / '_bundled')"
$ .venv/bin/python <that dir>/pip-25.2-py3-none-any.whl/pip --version
pip 25.2 from .../ensurepip/_bundled/pip-25.2-py3-none-any.whl/pip (python 3.13)
```

When an interpreter has neither pip nor a bundled wheel (Debian's `python3-minimal` removes `ensurepip`), the launcher should say so by name instead of reporting `No module named pip`.

The two installer test files passing in a pip-less virtual environment, and a test that pins the resolution order: installed pip first, bundled wheel second, a named error third.

### LINKED ISSUE #583: [bug] wiki_refine_draft with only a draft_id stamps synth_model and writes a refined_llm memo
### What happened

`wiki_refine_draft` (`mcp_server/handlers/wiki_refine.py::handler_refine`) accepts a call carrying nothing but `draft_id`. `title`, `lead`, `sections` and `frontmatter` are all optional and `update_draft` skips every field that is `None`, but `synth_model` defaults to `claude_refine_v1` and is always passed, so the update touches a row, returns `updated: true`, and `insert_memo` writes a `refined_llm` row into `wiki.memos`.

A call that refines nothing therefore leaves an audit trail saying a model refined the draft, and stamps the draft's `synth_model` with a model that did not write its content. Since #579 the tool is registered, so any MCP client can make that call.

Found by the review of PR #581 (https://github.com/cdeust/Cortex/pull/581#issuecomment-5692749644). The behavior predates that PR; exposing the tool is what makes it reachable.

### Expected behavior

A refinement carries content. When `title`, `lead`, `sections` and `frontmatter` are all absent, the handler returns an error naming what it expects and writes nothing: no `synth_model`, no memo.

A test that calls the handler with only a `draft_id` and asserts the draft row and `wiki.memos` are both unchanged, and that the response is an error.

cdeust:
## Summary

Two fixes the review of #581 and the local suite surfaced.

`scripts/launcher_pip.py` built every install command as `sys.executable -m pip`, so an interpreter without a pip module could not fill the plugin's private `deps/` at all, reporting `No module named pip`. A virtual environment made by `uv venv` or `python -m venv --without-pip` has no pip module by design. `pip_entry()` now resolves the installed module first, then the pip wheel CPython bundles for `ensurepip`, run in place from the wheel path so nothing is installed into the interpreter, and otherwise raises a `FileNotFoundError` naming that interpreter, which travels the `OSError` path both call sites already report (ADR-1067). `tests_py/scripts/test_installer_upgrade_single_version.py` seeds its fixture through the same resolution instead of hardcoding `-m pip`.

`mcp_server/handlers/wiki_refine.py` accepted a `wiki_refine_draft` call carrying nothing but a `draft_id`: `synth_model` always defaulted, so the draft was stamped with a model that wrote none of its content and a `refined_llm` memo recorded a refinement that never happened. Such a call is now refused whole, with no write to the draft and no memo (ADR-1068).

Closes #582
Closes #583

## Type of change

## Test plan

- New `tests_py/scripts/test_launcher_pip_entry.py` pins the resolution order: installed module preferred; bundled wheel used when no pip module is importable; that resolved command answering as pip when actually run; the interpreter named when neither exists; `_pip_command` built on the resolved entry. Four of the five fail on the previous code.
- New `test_a_refine_carrying_no_content_writes_nothing` in `tests_py/handlers/test_wiki_refine_confidence.py` asserts the draft's `synth_model` and `lead` are unchanged, the memo count for that draft is unchanged, and the response is an error. It fails on the previous handler.
- The 12 tests of `test_installer_rerun_keeps_entries.py` and `test_installer_upgrade_single_version.py` pass in this pip-less `uv venv`, where 11 of them errored in setup before.
- Full suite: 8675 passed, 8 skipped, no errors.
- `ruff check`, `ruff format --check`: clean. `pyright mcp_server/`: 0 errors. `scripts/check_craftsmanship.py --base origin/main`: OK. `scripts/check_doc_claims.py`: OK.

## Audit notes

- `bundled_pip_wheel()` locates `ensurepip` with `importlib.util.find_spec` rather than importing it, so a standard library without the module returns `None` instead of raising at import time.
- The bundled wheel's pip can be older than an installed pip. Installs are pinned and hash-checked (ADR-1059), so the resolver version selects no different distribution.

- [x] §2.2 Layer dependency direction preserved.
- [x] §4.1 / §4.2 No new size violation; the craftsmanship gate passes against origin/main.
- [x] §8 No numeric constant added.

## Breaking changes

None. A `wiki_refine_draft` call with only a `draft_id` used to return `updated: true` after writing a memo; it now returns an error. The tool has been registered only since #581 and the store holds one such memo, from a test runner in April 2026.

## Reviewer checklist

- [x] Documentation updated (ADR-1067, ADR-1068, the `wiki_refine_draft` tool description).


cdeust:
ZETETIC-REVIEW: APPROVE

## Summary
Both fixes are correct, mutation-tested, and match the PR's claims exactly. `launcher_pip.pip_entry()` resolves installed-pip → bundled-ensurepip-wheel → named `FileNotFoundError`, and both call sites (`launcher_deps_install.py::pip_install`, `install_requirements`) genuinely catch it via their existing `except OSError` handlers (`FileNotFoundError` is an `OSError` subclass) - no traceback, clean `False` return. `wiki_refine.py`'s `_CONTENT_FIELDS` refusal is correct for every shape I drove it against, including the empty-string/empty-list/empty-dict edge cases the task asked about. One CI job (`Test (SQLite backend)`) was cancelled on a 10-minute execution-time budget; evidence below indicates this is a general runner-slowness/timeout-budget issue this run, not a defect in the diff - flagged as a merge-blocking item to resolve with a rerun, separate from the code-review verdict.

## Move 0 - Ledger reconciliation and seen-defect check (§13.2 + §14)
- No embedded Completion Ledger in the PR body; independently re-enumerated the diff's paths myself (11 files: CHANGELOG, 2 ADR pairs (docs/ + wiki mirror), handler, launcher_pip.py, 3 test files, wiki/manifest.json). All accounted for; no unmapped branch/error-arm.
- Seen-defect rationalizations: none found in the diff or PR body. The reviewer checklist explicitly leaves "CI passes on latest commit" **unchecked** - the author did not claim green CI, so there is nothing here to refuse over.
- Verdict short-circuit: N/A, proceeding.

- Classification: **Medium**.
- Criterion: core packaging/installer logic used by every plugin install, plus a documented breaking change to a registered MCP tool contract (`wiki_refine_draft`); not auth/billing/crypto/migrations, not >400 lines, <2 distinct authors in 90d (`cdeust`/`Clement` are the same person per `git log`).
- Depth applied: full Moves 1-6.

## Layer check (Move 1)
| File | Layer | Imports added/changed | Verdict |
|---|---|---|---|
| `scripts/launcher_pip.py` | scripts (stdlib-only tooling) | `+import importlib.util` (stdlib) | OK - no layer crossed |
| `mcp_server/handlers/wiki_refine.py` | handlers | none added; existing `infrastructure.*` imports only | OK |

## SOLID audit (Move 2)
No violations. `pip_entry()`/`bundled_pip_wheel()` are new, single-purpose functions (resolution only); `_pip_command` was refactored to depend on the new entry point rather than growing a branch (this *is* the OCP-friendly shape - no new `if backend == X` added to the install path). The refusal in `handler_refine` is a guard clause, not a type-switch.

- New public symbols: `launcher_pip.bundled_pip_wheel()`, `launcher_pip.pip_entry()`. Both wired: `pip_entry()` is called from `_pip_command()` (still the sole caller of `_install`/`resolve`/`install_requirements`), and independently by `tests_py/scripts/test_installer_upgrade_single_version.py::_pip_entry()`.
- `_CONTENT_FIELDS` constant: wired into the new refusal branch in `handler_refine`; also referenced by the updated tool description string (`schema_refine["description"]`) - consistent.
- Contract change: `wiki_refine_draft` now refuses a `draft_id`-only call. PR body documents this as the only breaking change and cites the one historical caller (a test-runner memo, April 2026) - acceptable, no other callers in-repo (`grep -rl "wiki_refine_draft"` confirms no internal caller besides the handler/registry/tests).
- No dead code, no bare TODOs, no debug prints in the diff.

## Test adequacy (Move 4) - evidence, not assertion
Ran everything myself in the pip-less worktree venv (`.venv/bin/python3`, confirmed `ModuleNotFoundError: No module named 'pip'`, confirmed `ensurepip` bundles `pip-25.0.1-py3-none-any.whl` under uv's cpython-3.12.11):

1. `pip_entry()` resolution order, driven directly:
   `pip_entry()` → `['.../pip-refine/.venv/bin/python3', '.../ensurepip/_bundled/pip-25.0.1-py3-none-any.whl/pip']` (bundled-wheel branch, correct - no installed module).

2. Failure reaches existing reporting, no traceback. With both `find_spec("pip")` and `bundled_pip_wheel()` monkeypatched to fail-both:
   - `launcher_deps_install.pip_install(...)` → prints `[cortex-launcher] dependency resolution failed: ... has no pip module and its standard library bundles no pip wheel for ensurepip...` and returns `False`. Confirmed via `except (OSError, _cpu.CpuWheelError)` at `scripts/launcher_deps_install.py:230`.
   - `launcher_deps_install.install_requirements(...)` → same pattern via `except OSError` at `scripts/launcher_deps_install.py:253`. `FileNotFoundError` is an `OSError` subclass (stdlib), confirmed both catch it - no crash, no bare traceback.

3. Real install, not just `--version`. Built a wheel for a synthetic `probepkg==1.0`, wrote a hash-pinned `requirements.txt` with `--no-index --find-links <wheel dir>` (pip honors these as in-file directives, same trick `test_installer_upgrade_single_version.py` doesn't need since it drives `_pip_target` directly - I drove the higher-level `scripts/launcher_deps.py --requirement <file> <deps_dir>` entry point instead, per the task's ask), then ran it in the pip-less venv:
   `UV_NO_SYNC=1 uv run python scripts/launcher_deps.py --requirement requirements.txt depsout` → exit 0, `depsout/probepkg/__init__.py`, `depsout/probepkg-1.0.dist-info/{RECORD,WHEEL,METADATA,INSTALLER,REQUESTED}` all present. Genuine install via the bundled-wheel pip, in the pip-less interpreter.

4. `bundled_pip_wheel()` correctness:
   - `ensurepip` present, `_bundled` non-empty (the real case here): returns the sorted-last `pip-*.whl` - correct, and `sorted()` + `[-1]` picks the highest version string if ever more than one wheel is bundled (CPython historically ships exactly one, but the code degrades gracefully if that ever changes).
   - `ensurepip` present but `_bundled` empty/missing: `wheels` is `[]`, returns `None` - correct, falls through to the named error. (Not independently reproducible without patching a real stdlib, but the code path is a direct, exhaustively-tested glob-empty case identical to what `test_an_interpreter_without_either_is_named_in_the_error` monkeypatches.)
   - `ensurepip` absent (`find_spec` returns `None`, e.g. Debian `python3-minimal`): returns `None` immediately via the `spec is None` guard - correct.

5. Suite numbers match exactly. `UV_NO_SYNC=1 uv run pytest -q` in this worktree: `8675 passed, 8 skipped ... in 278.41s` - identical to the PR body's claim.

6. `_CONTENT_FIELDS` refusal - every shape driven directly against a real SQLite-backed draft (via the existing `sqlite_store` fixture):
   - `{"draft_id": N}` (all four absent) → refused, `updated: False`, error names the four fields, draft/memo unchanged (this is the PR's own new test, reran and confirmed green).
   - `{"draft_id": N, "lead": ""}` → **treated as content**: `updated: True`, `lead` written as `''`, a `refined_llm` memo IS inserted (memo count 1→2). This is correct per the code's own semantics: `args.get("lead")` is `""`, not `None`, so it is not "absent" - and `update_draft`'s own skip-on-`None` logic uses the identical predicate, so a client that explicitly sends `lead=""` is asking to clear the lead, a real edit, and the audit memo correctly records that a (trivial) edit happened. The MCP schema types `lead` as `"string"` with no `null` variant, so a client cannot send an explicit null to mean "omitted" - omission is the only way to signal "no opinion," and `args.get()` correctly returns `None` only for that case.
   - `{"draft_id": N, "sections": []}` → not refused by `_CONTENT_FIELDS` (an empty list is not `None`), but caught immediately after by the existing `_validate_against_contract` check (`required section missing: 'Context'`, etc. for an ADR-kind draft) - so it still can't silently "succeed while doing nothing"; it fails validation instead. No write occurs either way.
   - `{"draft_id": N, "frontmatter": {}}` → same as `lead=""`: treated as content, `updated: True`, memo written. Symmetric and consistent with `update_draft`'s skip-on-`None`.
   - Net: the `is None` test is correct for ADR-1068's stated scope ("all absent") and does not misclassify any of the three edge shapes the task raised - they are all legitimate explicit edits, not silently-nothing refinements. **Non-blocking gap**: none of these three edge shapes has a test in this PR (only the all-`None` case is tested); worth a follow-up test for `lead=""`/`frontmatter={}` since the "counts as content" behavior is a real design decision someone could accidentally invert later.

`pip_entry()` (16 lines), `bundled_pip_wheel()` (11 lines), the refusal block in `handler_refine` (8 lines) - all well under the 40-line/300-line thresholds. `scripts/check_craftsmanship.py --base origin/main` → `OK`. No over-engineering smells.

- No new I/O on user-controlled input beyond what already existed (pip install from a hash-pinned requirements file, unchanged trust model per ADR-1059).
- `bundled_pip_wheel()` uses `find_spec` (no import side effect) - correctly avoids importing `ensurepip` (which has installation side effects) just to check for its presence.
- Commit hygiene: single conventional commit (`fix(launcher,wiki): ...`), scoped to the two linked issues, no unrelated changes, no secrets.

## Gates run (this worktree, `.venv` synced without pip per the task setup)
- `UV_NO_SYNC=1 uv run ruff check` → All checks passed.
- `UV_NO_SYNC=1 uv run ruff format --check` → 1458 files already formatted.
- `UV_NO_SYNC=1 uv run python scripts/check_craftsmanship.py --base origin/main` → OK.
- `UV_NO_SYNC=1 uv run python scripts/check_doc_claims.py` → OK (1 declared exemption, pre-existing, CONTRIBUTING.md:36).
- `.venv/bin/python -m pyright mcp_server/` → 0 errors, 0 warnings, 0 informations.
- `UV_NO_SYNC=1 uv run pytest -q` → **8675 passed, 8 skipped**, 459 subtests passed in 278.41s - matches the PR body exactly.
- `git status --short` after all of the above → clean.

## Mutation testing (Move 4 strength check, §12) - scratch copies, worktree untouched
Per the task's "scratch copy" instruction, mutated file *copies* under the scratchpad (never the worktree itself - `git status --short` stayed clean throughout, confirmed after every run):
1. Dropped the bundled-wheel branch from `pip_entry()` (falls straight to the `FileNotFoundError` when no installed module): loaded the mutant as `sys.modules["launcher_pip"]` and ran `tests_py/scripts/test_launcher_pip_entry.py` → **1 failed** (`test_without_a_pip_module_the_bundled_wheel_is_used`), 4 passed. Mutant killed.
2. Changed the refusal condition `if all(...)` to `if False` in `wiki_refine.py`: loaded the mutant as `sys.modules["mcp_server.handlers.wiki_refine"]` via a `pytest_configure` plugin hook (needed so the test-isolation conftest guard runs first) and ran `tests_py/handlers/test_wiki_refine_confidence.py` → **1 failed** (`test_a_refine_carrying_no_content_writes_nothing`, `assert True is False`), 3 passed. Mutant killed.
Both fixes are mutation-strong, not just line-covered. `git status --short` in the worktree: clean before, during, and after.

## Doc/ADR/CHANGELOG accuracy
- ADR-1067's Homebrew claim ("Python 3.14.4, which carries pip") verified: local `/opt/homebrew/bin/python3` is 3.14.4 and `import pip` → `pip.__version__ == '26.0.1'`.
- ADR-1067's `ensurepip/_bundled` claim verified directly on the CI-adjacent uv-managed interpreter (see Move 4 §1 above).
- ADR-1068 references "ADR-1066 registered `wiki_refine_draft`" - cross-checked against `git log`: PR #579 (merged via #581, ADR-1066) is exactly "expose the Path B draft tools over MCP." Consistent, not a contradiction with the commit message's "#579" framing.
- CHANGELOG entries for #582/#583 match the code and the ADRs; ordering under "Fixed" is chronological-by-relevance (newest first), consistent with the existing #578/#573 entries below them.

## Issues
### Blocking (merge-gating, not code-defect)
- CI job `Test (SQLite backend)` was cancelled (exceeded the 10-minute job budget; `gh api repos/cdeust/Cortex/actions/jobs/104687037941` → `conclusion: cancelled`, annotation "The job has exceeded the maximum execution time of 10m0s"). Its own pytest step took 505.57s (8373 passed, 141 skipped) vs. 224.74s on the latest `main` run (8367 passed, 141 skipped) for the same suite minus 6 tests - a 2.25x slowdown I cannot attribute to this diff: `Test (Python 3.10)` (7m31s) and `Test (Python 3.12)` (8m49s) on this same PR run were also markedly slower than their typical ~5-6 min baseline while still completing, which points to general runner contention on this specific workflow run rather than a hang or regression introduced by the 6 new tests (each of which is sub-second in isolation, confirmed locally). Per this project's own rule (CI green + posted review = merge), **do not merge on this run's red check** - rerun CI and confirm `Test (SQLite backend)` completes within budget before merging. If the slowdown reproduces on a clean rerun, that would warrant reopening this as a real defect; on the evidence gathered here it looks like transient infra noise, not a defect in the diff.

### Non-blocking
- `tests_py/handlers/test_wiki_refine_confidence.py` - no test exercises `lead=""` or `frontmatter={}` as content (both correctly `updated: True` per the code's design, verified manually above); a follow-up test pinning that "explicit-empty counts as content" behavior would protect against someone later "fixing" `_CONTENT_FIELDS`'s falsy-vs-None distinction by accident.

## Hand-offs
None. No root-cause-fix, structural-decomposition, formal-correctness, cargo-cult, threat-modeling, or performance-claim concerns beyond this review's scope.

## Verdict
**APPROVE** at the code-review level - both fixes are correct, minimal, mutation-tested, and match every claim in the PR body and ADRs that I could independently verify. **Do not merge on the current CI run**: rerun and confirm `Test (SQLite backend)` passes within its time budget first (see Blocking item above), per this project's CI-is-the-authority merge gate.

PR #600, opened 2026-09-16
fix: align Claude and Codex backend selection and verify shared handoff

cdeust:
## Summary

Direct Cortex MCP startup now applies the saved backend selection before settings-dependent imports, matching the Claude launcher. Previously, a saved SQLite selection could be honoured by Claude while the Codex console entry point selected a reachable PostgreSQL server, sending the two hosts to different stores.

Adds independent-process tests for shared SQLite persistence and project ADR handoff, plus documentation of the common storage/root contract. The Codex plugin retains its existing ten-tool lean profile; ADR authoring requires an explicitly configured full-profile server.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Supporting tests and documentation

## Test plan

- New tests: **12 passed** (bootstrap selection, real MCP project ADR calls, independent SQLite processes).
- Related existing tests: **47 passed** (backend marker, launcher resolution, tool profiles, ADR identity).
- Regression baseline: **4 of 9 bootstrap cases failed before the fix**, all nine pass afterward.
- Ruff lint and formatting, explicit five-file craftsmanship gate, and `git diff --check`: passed.

Commands, using the repository Python environment and `CORTEX_MEMORY_STORE_BACKEND=sqlite`:

```sh
python -m pytest -q tests_py/test_shared_backend_bootstrap.py tests_py/test_shared_host_decisions.py tests_py/infrastructure/test_shared_host_store.py
python -m pytest -q tests_py/infrastructure/test_backend_marker.py tests_py/scripts/test_launcher_resolution.py tests_py/test_tool_profiles.py tests_py/handlers/test_wiki_adr_identity.py
```

Tests cover exact ADR retrieval, denied lean authoring, explicitly linked follow-up decisions, concurrent ADR submissions, separation between explicit project roots, visibility of acknowledged SQLite writes, and persistence after reopening. Child processes have bounded protocol waits and cleanup.

The full suite has not been run locally. These are local macOS/POSIX results, not Windows coverage. Installed Claude/Codex clients, semantic retrieval quality, PostgreSQL handoff, global ADR-to-database pointer integration, and automatic host capture remain outside this verification. Concurrent submissions do not prove every possible race absent. No mutation campaign was run.

## Audit notes

- Independent engineering review found missing subprocess deadlines/cleanup; fixed and re-reviewed with an APPROVE verdict for the bounded scope.
- Formal Zetetic team review: pending, arranged by the owner.
[...]

PR #601, opened 2026-09-16
chore(git): ignore .Codex/, where Codex puts its worktrees

cdeust:
## Summary

The project's standing rule puts every git worktree at `.claude/worktrees/<name>/`, a path `.gitignore` already covers. Codex uses `.Codex/worktrees/` instead, and nothing ignored it: reviewing PR #600 found a sixty-entry checkout sitting at `.Codex/worktrees/shared-host-decisions`, tracked by nothing and one `git add -A` away from being committed into the repository.

This ignores `.Codex/`. It does not move the worktree, which is the owner's call and unrelated to whether the path can be committed by accident.

## Type of change

## Test plan

- Created `.Codex/worktrees/probe/file.txt` in this worktree: `git status --porcelain --untracked-files=all` lists zero `.Codex` entries, and `git check-ignore -v` names `.gitignore:66` as the matching rule. Removed the probe afterwards.
- No source file changed, so the suite is unaffected.

## Audit notes

- The rule sits next to the existing `.claude/*` block, with the observation that produced it, so the next reader knows which tool creates that path and why it is ignored rather than moved.

## Breaking changes

None. A path that was never tracked stops being offered as untracked.

## Reviewer checklist

- [ ] CHANGELOG.md updated: not applicable, this changes no behaviour of the shipped package.


cdeust:
ZETETIC-REVIEW: APPROVE

Reviewed 8ccdb27a from a detached worktree at `.claude/worktrees/review-601`, base `origin/main` 0fc6bd1a. Nothing edited, nothing pushed.

## Stakes (Move 7)

Low. One repository-config file, five added lines, no source, no public surface, no security path, no test impact. Moves 1, 3 and 6 still run, since they run at every stakes level. Moves 2, 4 and 5 are informal here because there is no unit to audit.

## Move 0, ledger and seen-defect check

The diff has exactly one path: a `.gitignore` pattern. I re-enumerated it against the PR body and found no unmapped branch and no unclaimed effect.

One scope statement in the body ("It does not move the worktree, which is the owner's call") reads like a deferral, so I checked whether it dismisses a defect this diff touches. It does not. Where Codex writes its worktrees is host-tool configuration outside the repository tree, unreachable from this diff and from every check it ran, and the author names the deferral rather than hiding it. Not a section 14 bypass. There is a related note below under Non-blocking.

## Verification, re-run independently rather than taken from the PR body

1. The rule matches what it claims. I created `.Codex/worktrees/probe/file.txt` in the worktree. `git status --porcelain --untracked-files=all` returned 0 lines in total, with no `.Codex` entry. `git check-ignore -v .Codex/worktrees/probe/file.txt` returned `.gitignore:66:.Codex/`. I then removed the probe and the tree came back clean. The commit message's claim of `.gitignore:66` is exact: the pattern sits on line 66.

2. The rule is not broader than intended. `git ls-files | grep -i codex` returns 11 tracked paths (`docs/codex-plugin.md`, the `plugins/hypermnesia-mcp-codex/` tree including `.codex-plugin/plugin.json` and `.codexignore`, and `tests_py/scripts/test_codex_plugin_contract.py`). I ran `git check-ignore -v` over each one and none is ignored, so no tracked path changes status. The `.claude/` behaviour is untouched: `.claude/worktrees/x` and `.claude/other.json` still match `.gitignore:60:.claude/*`, and `.claude/settings.json` is still not ignored, so the `!` exception on line 61 holds. The new rule sits after that negation and matches a disjoint prefix, so it cannot shadow it.

3. The comment is four consecutive comment lines, 62 to 65. The longest comment run anywhere in `.gitignore` is 4, set by the pre-existing `.claude` block on lines 56 to 59, well under the decision gate's eight. The content is a measured observation rather than a claim, and I confirmed the observation itself: `.Codex/worktrees/shared-host-decisions` exists in the primary checkout with 61 paths at depth 3 or less, PR #600 is open and is the review that surfaced it, and on `main` without this rule the directory appears as `?? .Codex/`. The "one `git add -A` away" statement is literally true.

4. Both gates pass on the branch, after `uv sync --no-default-groups --extra dev --extra sqlite` and with `UV_NO_SYNC=1 uv run`. `scripts/check_doc_claims.py` printed `doc claims OK (1 declared not-a-claim exemption(s))` and exited 0. `scripts/check_craftsmanship.py --base origin/main` printed `Craftsmanship gate: OK` and exited 0.

5. `gh pr checks 601` reports 24 pass and 4 skipping (Tag a release, both Docker Build images, and the scheduled fuzz batch, all correctly inert for a non-source change). Nothing failing, nothing pending.

## Moves 1, 3 and 6

Layer boundary: no import and no module, so nothing applies. Wiring and contract drift: no symbol added, renamed or deleted, and no caller to update; no commented-out code, no untracked TODO, no debug statement. Security and hygiene: no secret, no PII, no binary, no generated file; conventional commit (`chore(git):`), one logical change, 5 insertions and 0 deletions, so no scope creep. The change can only remove a path from `git add` reach, so it cannot cause anything to be committed.

## Adversarial pass, four apertures, all examined

Over-fit: the pattern is unanchored, so it matches a `.Codex` directory at any depth rather than only the repository root. I verified this is harmless, since nothing tracked matches.

Missed cases: the intended path `.Codex/worktrees/**` is covered at every depth, and an unanchored pattern cannot under-match relative to an anchored one.

Robustness: an already-tracked file under `.Codex/` would stay tracked despite the rule, but `git ls-files` shows none exists, so there is no silent-divergence case.

Test adequacy: there is no automated test, and none is warranted. The rule's behaviour is a single `git check-ignore` fact, which I reproduced rather than trusted. A `.gitignore` regression test here would be ceremony.

No blocking finding at any aperture.

## Non-blocking

`.gitignore:66` is unanchored. `/.Codex/` would express the actual intent, a root-level tool directory, and would not silently swallow a stray `.Codex` nested deeper in the tree. Nothing today is affected; this is only about the rule saying what it means.

`.gitignore:66` carries a case-fold asymmetry worth knowing about. This clone has `core.ignorecase=true`, the APFS default, under which `.Codex/` also matches a lowercase `.codex/`; I confirmed that it does. On Linux CI it would not. If Codex ever emits `.codex/`, the rule works on the author's machine and not in CI. That is not a defect of this diff, only a fact to hold before relying on the rule for a lowercase path.

The deferral "it does not move the worktree" is a sound scope call, but the underlying condition, a worktree living outside the path the project rule names, is now invisible to `git status` rather than resolved. An issue would keep it tracked rather than merely silenced.

## Verdict

APPROVE. The diff does what it says, its own commit message is reproducible down to the line number, it widens nothing that is tracked, `.claude/` semantics are intact, both gates pass on the branch and CI is green.

PR #608, opened 2026-09-17
feat(hooks): normalise Codex hook events into what the lifecycle hooks read (#608)

cdeust:
## Summary

PR 2 of the "Cortex behaves the same under Codex as under Claude Code"
wave. PR 1 (#605) added `mcp_server/hooks/entry.py`, a console entry point
that runs one of the eleven lifecycle hook modules from the wheel, but
dispatched whatever stdin carried unchanged. This PR normalises a Codex
hook event into the Claude-shaped event the modules already read, then
runs the module once per derived event.

- `mcp_server/hooks/host_patch.py`: a fail-closed `apply_patch` parser,
  ported from the same owner's zetetic-team-subagents
  `hooks/lib/host_events.py` (commit 4c2bc9b, PR #139, MIT licensed).
  Produces an Edit/Write-shaped event per file operation; refuses
  ambiguous or missing preimage context rather than approving content
  different from the prospective edit.
- `mcp_server/hooks/host_event.py`: `normalize_event(event)` dispatches
  Claude events through unchanged, expands `apply_patch` via
  `host_patch`, turns `exec_command`/`shell_command` into `Bash` with
  `workdir` resolved against `cwd`, and maps `SubagentStart`'s
  `agent_type` to `agent_name`.
- `mcp_server/hooks/entry.py`: reads stdin once, normalises it, sets
  `CLAUDE_PROJECT_ROOT` from the event's `cwd` only when unset, and runs
  the hook module once per derived event. For a `PreToolUse` event the
  first non-zero exit wins and stops the loop.

## Test plan

- [x] `tests_py/hooks/test_host_patch.py` (ported zetetic fixtures: Add
      File, Update with Move to, Delete File, End of File, non-exact and
      ambiguous preimage refusal)
- [x] `tests_py/hooks/test_host_event.py` (Claude pass-through pins,
      shell workdir resolution, SubagentStart mapping,
      PreCompact/PostCompact pass-through, apply_patch dispatch)
- [x] `tests_py/hooks/test_entry.py` additions: `apply_patch` end to end
      through `preemptive_context`/`pipeline_impact_bump` matching the
      equivalent Claude Edit event's stdout; `decision_gate` blocking an
      `apply_patch` Add File the same way it blocks the equivalent Claude
      Write; malformed patch exits 2 on PreToolUse, 1 with a stderr line
      on PostToolUse
- [x] `uvx ruff@0.16.6 check` / `format --check`: clean
- [x] pyright on changed files: 0 errors
- [x] `python scripts/check_craftsmanship.py --base origin/main`: OK
- [x] `python scripts/check_doc_claims.py`: OK
- [x] Full `pytest -q`: 8839 passed, 8 skipped, 459 subtests passed in 387.17s


cdeust:

Reviewed 0f7ea00f (rebased on origin/main 4bd7e7f0, merge-base equals main) in a read-only worktree at `.claude/worktrees/review-608` with the repo interpreter. Stakes: high. This is the path every Codex hook event takes, and it produces the derived events the edit gates block on.

The parser is the strong part of this change set. The parity and fail-closed properties I could think of all hold under measurement. What is not carried by any test is the dispatch loop in `entry.main()`, and that is what I am blocking on.

## Required changes

### The PreToolUse stop-on-first-block and the multi-event loop are uncovered

`entry.main()` gained two behaviours whose postconditions are stated in its own docstring: `for a PreToolUse event the first non-zero exit among the derived events wins and stops further dispatch (an apply_patch with several file operations must not apply the second once the first is blocked)`, and running the module once per derived event. Neither has a test. I deleted the break and ran everything:

```
$ python3 -  # remove the two lines `if hook_event_name == PRE_TOOL_USE and code != 0:` / `break`
$ .venv/bin/python -m pytest -q -p no:randomly
8852 passed, 8 skipped, 2 warnings, 459 subtests passed in 423.50s (0:07:03)
```

Identical to the clean run on this tip. The mutant lives through the whole suite. A safety claim that no test would notice being deleted is not a safety claim yet. Please add a test that dispatches a two-operation `apply_patch` on `PreToolUse` where the first operation is blocked, and asserts the second module run never happened (patching `host_dispatch.run_event` and counting calls is enough; the subprocess route cannot see the count).

### CLAUDE_PROJECT_ROOT precedence is uncovered in both arms

`entry.py:114` reads `if cwd and "CLAUDE_PROJECT_ROOT" not in os.environ:`, and the docstring explains why the existing value must win (the Claude launcher's cwd is the plugin root). `grep -rn CLAUDE_PROJECT_ROOT tests_py/` returns hits only in `test_session_start.py` and `test_pipeline_lazy_imports.py`, none reaching `entry`. The same full run above also carried that condition weakened to `if cwd:`, which inverts the precedence, and the suite stayed at 8852 passed. Inverted, `session_start.py:919`, `pipeline_impact_bump.py:195` and `post_commit_reindex.py:255` would all resolve the wrong project root under Claude Code, which is the host this must not regress. Two assertions cover it: preset value kept, unset value filled from the event `cwd`.

I verified both arms behave correctly today by driving `entry.main()` in process with `run_event` replaced:

```
(d) CLAUDE_PROJECT_ROOT seen by run: /var/folders/.../tmp_wr6i8mt (= event cwd: True )
preset kept: /preset/elsewhere
```

So the code is right. It is the suite that does not hold it.

## One divergence from the port source, to confirm or restore

`hooks/lib/host_events.py` at 4c2bc9b resolves a shell event's directory as `(base / (tin.get("workdir") or tin.get("cwd") or base)).resolve()`. `host_event._shell_event` keeps `workdir` and drops the `tool_input["cwd"]` fallback. Measured on this tip:

```
exec_command {"cmd":"git commit -m x","workdir":"sub"} -> cwd /p/sub
exec_command {"cmd":"ls","cwd":"sub"}                  -> cwd /p
```

If Codex ever puts the directory under `cwd` inside `tool_input`, `post_commit_reindex` reindexes the parent instead of the repository that was committed in, silently. The module docstring claims the field set was verified against learn.chatgpt.com/docs/hooks and against that same upstream file, so either the fallback comes back with a test, or the docstring says in one sentence why `workdir` is the whole contract and the upstream fallback is dead weight. Right now the divergence is invisible to a reader of either file.

## What I measured and what held

Port against its source. Function by function, `_resolve`/`_path`, `_read_preimage`/`_read`, `_match_unique`/`_match`, `_read_hunk`/`_hunk`, `_apply_hunks`/`_updated`, `_operation` and the envelope walk in `patch_events`/`_patch_events` are semantically identical, including the `_MIN_ENVELOPE_LINES = 3` floor now sourced in the fix commit. The two intentional differences are the `tool_input["cwd"]` drop above and the decision not to rewrite `file_path` on Claude `Edit`/`Write` events, which is what makes the pass-through pins exact.

Fail-closed properties, run rather than read:

```
ambiguous       => REFUSED: Patch preimage is ambiguous; exact unique context required
missing         => REFUSED: Patch preimage is missing; exact unique context required
end-of-file     => OK  Edit old "a\nb\na\n" new "a\nb\nA\n"   (anchored on the final line, not the first match)
move            => OK  Edit mv.txt -> "" , Write moved.txt "bye\n"
same-file-twice => OK  Write new.txt "one\ntwo\n" , Edit new.txt "one\ntwo\n" -> "one\nTWO\n"  (prospective state, not disk)
relative-subdir => OK  file_path <cwd>/sub/deep.txt
delete-missing  => REFUSED: Cannot read patch preimage .../nosuch.txt
files on disk unchanged: ['amb.txt', 'eof.txt', 'mv.txt']
```

A path escaping `cwd` resolves outside the project with no containment check: `*** Add File: ../../etc/x` yields `file_path` at `/private/var/folders/0c/etc/x`. That matches the upstream parser and, more to the point, matches what a Claude `Write` to the same absolute path already produces, so it opens no hole in the gates. I checked the gate rather than assuming it:

```
apply_patch escaping cwd -> rc 2, [decision-gate] BLOCKED: .../outside.py:1 would add 9 consecutive comment lines
claude Write same target -> rc 2, [decision-gate] BLOCKED: .../outside2.py:1 would add 9 consecutive comment lines
```

Pass-through pins. Fifteen Claude event shapes covering all eleven wired modules (SessionStart, UserPromptSubmit, PreToolUse Write and Edit, PostToolUse Bash/Read/Write, Stop, SessionEnd, PreCompact, PostCompact, SubagentStart, SubagentStop, Notification): `normalize_event(e) == [e]` JSON-equal in every case, with the input object unmutated. FAILURES: 0.

End to end through `mcp_server.hooks.entry` with isolated `HOME`/`CORTEX_CLAUDE_DIR` and the sqlite backend:

```
(a) preemptive_context   apply_patch rc=0 stdout '' | claude Write rc=0 stdout ''   EQUAL
(a) pipeline_impact_bump apply_patch rc=0 stdout '' | claude Write rc=0 stdout ''   EQUAL
(b) decision_gate  apply_patch rc=2 | equivalent Write rc=2, both "[decision-gate] BLOCKED: ... 12 consecutive comment lines"
(c) malformed PreToolUse  rc=2, stderr "[hypermnesia-mcp-hook] mcp_server.hooks.decision_gate: Invalid Begin Patch / End Patch envelope"
(c) malformed PostToolUse rc=1, stderr "[hypermnesia-mcp-hook] mcp_server.hooks.post_tool_capture: Invalid Begin Patch / End Patch envelope"
(d) exec_command workdir "sub" -> derived cwd <cwd>/sub  (match: True)
(e) two-operation patch through post_tool_capture -> module ran twice (two "[cortex-post-tool-capture] skip Write" lines vs one for a single-operation patch), rc 0
```

A second check on the gate side: a two-operation patch whose first file is clean and whose second adds nine comment lines exits 2 and names the second file, so the loop really submits each derived event to the gate rather than only the first.

Edge cases from the diff. `normalize_event` returning an empty list would make `entry.main()` exit 0 with nothing dispatched; I could not reach that state, since `patch_events` either raises or yields at least one event for any envelope that passes the three-line floor, so it stays a latent shape rather than a bug. `run_event` replaces `sys.stdin` per run and never restores it, which is harmless because each run installs its own `io.StringIO` and the process exits after the loop. `int(exc.code)` would raise on a string exit code; all eleven modules exit with integers (`decision_gate.main` and `no_deps_gate.main` are typed `-> int`), so it holds today.

For a `PostToolUse` event with several derived events the last run's code wins, not the worst: with a first run at 3 and a second at 0, `main()` exits 0. That is what the docstring says it does, and PostToolUse exit codes are advisory, so I am not blocking on it, but `max` would cost one line and would stop a first-operation failure disappearing behind a clean second one.

Gates on 0f7ea00f, run locally:

```
$ .venv/bin/python scripts/check_craftsmanship.py --base origin/main
Craftsmanship gate: OK
$ uvx ruff@0.16.6 check           -> All checks passed!
$ uvx ruff@0.16.6 format --check  -> 3381 files already formatted
$ .venv/bin/python -m pyright <the five modules>  -> errorCount 0, warningCount 0
$ .venv/bin/python scripts/check_doc_claims.py    -> doc claims OK (1 declared not-a-claim exemption)
$ .venv/bin/python -m pytest tests_py/hooks tests_py/scripts -q -p no:randomly
1543 passed, 372 subtests passed in 106.16s
```

`tests_py/scripts` (1006 passed on the previous head, folded into the 1543 above) still passes with the new modules present; none of `host_patch`, `host_event`, `host_event_errors` or `host_dispatch` carries a `__main__` block, so the composition-root wiring test has nothing new to wire. The craftsmanship numbers after the split: `test_entry.py` 159 lines, `test_entry_apply_patch.py` 165, `host_patch.py` 222, all under the 300 cap, and the MIT header in the `host_patch` docstring is read as a docstring by the decision gate rather than as a comment block.

Full suite on the previous head e527fced: 8839 passed, 8 skipped, 459 subtests in 658.57s. On this tip the two mutation runs above both reported 8852 passed, 8 skipped, matching the executor's clean number for the tip.

CI on 0f7ea00f: 24 pass, 4 skipping, 0 fail (Craftsmanship Gate pass 16s, Type Check pass 1m33s, Test 3.10/3.11/3.12/3.13, SQLite and Windows SQLite all pass).

The two test additions are about thirty lines against `host_dispatch.run_event` and `os.environ`, and they do not touch the parser, which needs nothing.

cdeust:
ZETETIC-REVIEW: APPROVE

Round 2 on 7e24acfb (rebased onto origin/main 19b78dfc, merge-base equals main). `git range-diff 4bd7e7f0..2e099f21 19b78dfc..7e24acfb` reports all three commits as `=`, so the rebase carried no content change and the measurements I ran on 2e099f21 apply unchanged. Fresh worktree, repo interpreter, read-only.

## The two blocking items are closed, measured by re-applying the mutations

I re-applied each mutation and ran `tests_py/hooks` to see which named test notices.

Removing the `break` from `host_dispatch.run_all`:

```
FAILED tests_py/hooks/test_entry_apply_patch.py::test_pretooluse_stops_after_first_blocked_operation
FAILED tests_py/hooks/test_host_dispatch.py::test_run_all_pretooluse_stops_on_first_nonzero
2 failed, 544 passed, 13 subtests passed in 253.16s
```

Weakening `if cwd and _PROJECT_ROOT_VAR not in environ:` to `if cwd:`:

```
FAILED tests_py/hooks/test_host_dispatch.py::test_apply_project_root_keeps_preset_value
1 failed, 545 passed, 13 subtests passed in 58.81s
```

I also mutated the two changes you made on the non-blocking items, since untested fixes to a review comment are the same defect in a new place. Replacing `code = max(code, run_code)` with `code = run_code`:

```
FAILED tests_py/hooks/test_entry_apply_patch.py::test_posttooluse_exit_code_is_the_worst_not_the_last
FAILED tests_py/hooks/test_host_dispatch.py::test_run_all_posttooluse_runs_all_and_takes_the_worst
2 failed, 544 passed, 13 subtests passed in 68.92s
```

Dropping the restored `or tool_input.get("cwd")` from `_shell_event`:

```
FAILED tests_py/hooks/test_host_event.py::test_shell_tool_falls_back_to_tool_input_cwd_when_no_workdir
1 failed, 545 passed, 13 subtests passed in 86.39s
```

Four mutants, four killed, each by a test whose name says what it protects. `apply_project_root` covers all three arms (`fills_unset_value_from_event_cwd`, `keeps_preset_value`, `no_op_on_falsy_cwd`), and `test_posttooluse_runs_every_derived_operation` pins the multi-event dispatch itself.

The subprocess tests are the part I want to single out. `_two_blocked_add_files_patch_event` builds two Add File operations that each independently trip the gate, so counting `[decision-gate] BLOCKED` in stderr counts how many times the module actually ran. That is an observable consequence rather than a call-count on a patched function, and it is why the PreToolUse and PostToolUse arms can be told apart at all. I reproduced it outside the suite on this tip:

```
PreToolUse   rc 2  BLOCKED count 1
PostToolUse  rc 2  BLOCKED count 2
PostToolUse blocked-then-benign  rc 2  BLOCKED count 1
malformed Pre rc 2 | malformed Post rc 1
```

The third line is the `max` change doing its job: the block is on the first operation, the second run exits 0, and the process still exits 2.

## The shell fallback matches the upstream order

```
exec_command {"cmd":"ls","workdir":"sub"}              -> cwd /p/sub
exec_command {"cmd":"ls","cwd":"sub"}                  -> cwd /p/sub
exec_command {"cmd":"ls","workdir":"w","cwd":"c"}      -> cwd /p/w
```

`workdir` wins over `cwd`, which is the `tin.get("workdir") or tin.get("cwd")` order in zetetic-team-subagents `hooks/lib/host_events.py` at 4c2bc9b. The divergence I raised is gone, and the docstring now names the fallback instead of leaving a reader to diff the two files.

## Regressions I re-checked rather than assumed

The extraction moved real code out of `entry.main()`, so the round-1 properties were re-run on this tip rather than carried over.

```
ambiguous       => REFUSED: Patch preimage is ambiguous; exact unique context required
missing         => REFUSED: Patch preimage is missing; exact unique context required
end-of-file     => OK  Edit "a\nb\na\n" -> "a\nb\nA\n"
move            => OK  Edit mv.txt -> "" , Write moved.txt "bye\n"
same-file-twice => OK  Write new.txt "one\ntwo\n" , Edit -> "one\nTWO\n"   (prospective state)
delete-missing  => REFUSED: Cannot read patch preimage .../nosuch.txt
files on disk unchanged: ['amb.txt', 'eof.txt', 'mv.txt']
passthrough pins: 15   FAILURES: 0
```

`entry.main()` is now four statements after the wiring, and `run_all` is twelve lines with both arms visible at once, which is a better shape than the loop it replaced: the PreToolUse rule and the PostToolUse rule no longer share a mutable accumulator whose meaning changes with the event name.

Gates on 7e24acfb, run locally:

```
$ .venv/bin/python scripts/check_craftsmanship.py --base origin/main   -> Craftsmanship gate: OK
$ uvx ruff@0.16.6 check                                                 -> All checks passed!
$ uvx ruff@0.16.6 format --check                                        -> 3385 files already formatted
$ .venv/bin/python -m pyright <the five modules>                        -> errorCount 0, warningCount 0
$ .venv/bin/python scripts/check_doc_claims.py                          -> doc claims OK (1 declared exemption)
$ .venv/bin/python -m pytest tests_py/hooks tests_py/scripts -q -p no:randomly
1552 passed, 372 subtests passed in 106.65s
$ .venv/bin/python -m pytest -q -p no:randomly          (identical content, 2e099f21)
8861 passed, 8 skipped, 2 warnings, 459 subtests passed in 652.98s (0:10:52)
```

8861 matches your number exactly, and it is 9 above the 8852 of the previous head, which is the nine tests added here. Sizes stay under the caps: `host_dispatch.py` 108, `entry.py` 121, `host_event.py` 88, `host_patch.py` 222, `test_host_dispatch.py` 66, `test_entry_apply_patch.py` 236.

CI on 7e24acfb: 24 pass, 4 skipping, 0 fail.

One thing left standing from round 1, still not worth a change: `run_all` returns 0 if `payloads` is ever empty, and its docstring now states the non-empty precondition, which is the right place for it given `patch_events` either raises or yields at least one event for any envelope past the three-line floor.

PR #610, opened 2026-09-17
feat(transcripts): read Codex rollouts, write the session log before consolidating

cdeust:
## Summary

PR 3 of the "Cortex behaves the same under Codex as under Claude Code" wave. Two independent fixes.

(A) `mcp_server/infrastructure/transcript_activity.py` read only Claude Code's `{"type": "assistant", ...}` transcript shape, so a Codex SessionEnd fed procedural mining nothing: a Codex rollout's tool calls are `response_item` records of payload type `function_call` or `custom_tool_call`, not `assistant` message content. `transcript_activity` now peeks a transcript's first record and dispatches to a new `mcp_server/infrastructure/transcript_rollout.rollout_activity` when it is a `session_meta` record; every other transcript, including one whose first line cannot be read or parsed, is read unchanged by the existing Claude reader (pinned against every existing Claude fixture). ADR-1081.

(B) `mcp_server/hooks/session_lifecycle.py` ran consolidation in-process after the session-log write, which cost nothing extra under Claude Code (SessionEnd has no enforced timeout there) but risked the hook under Codex, where SessionEnd has a 1 s soft / 3 s hard timeout well under a consolidation cycle's cost. Consolidation now runs in a subprocess this module re-invokes with `--consolidate <mode>`, spawned detached (`Popen`, `start_new_session=True`, not waited on) after the log write, at the same light/standard/full depth the turn count already gated it to. Deliberately not routed through the existing `consolidate_background` launcher: that module's cycle and TTL stamp are SessionStart's periodic sweep, a different concern whose stamp would get tangled with a per-session cycle's. ADR-1082.

## Test plan

- [x] Gate proof (A): `tests_py/infrastructure/test_transcript_rollout.py::test_a_rollout_yields_its_tool_sequence_and_turn_count` fails against unmodified `transcript_activity.py` (`tool_sequence == []`), passes after the fix.
- [x] Gate proof (B): `tests_py/hooks/test_session_lifecycle.py::TestSpawnConsolidation::test_the_session_log_row_exists_even_when_the_launcher_fails` fails to collect against unmodified `session_lifecycle.py` (`_consolidation_mode` does not exist there), passes after the fix.
- [x] `uvx ruff@0.16.6 check .` / `format --check .` — clean.
- [x] `pyright` on the changed files (repo `.venv`, pyright 1.1.411) — 0 errors, 0 warnings.
- [x] `python scripts/check_craftsmanship.py --base origin/main` — OK, net −2 violations vs `origin/main` (baseline regenerated, no new debt).
- [x] `python scripts/check_doc_claims.py` — OK.
- [x] `python scripts/check_project_wiki.py` — OK.
- [x] Full `pytest -q`: 8893 passed, 8 skipped, 459 subtests passed in 495.71s (0:08:15).


cdeust:
ZETETIC-REVIEW: APPROVE

## Move 0, ledger reconciliation and seen-defect check
Every branch and degraded path in the diff is accounted for: the session_meta dispatch, the Claude-reader fallback (missing file, malformed first line, JSON-array-not-JSONL), the rollout reader's per-record skip cases, the spawn-failure non-fatal path, and the unrecognized-mode default to light. No rationalization phrase such as "pre-existing" or "unrelated" or "out of scope" appears in the PR body or the diff's own comments. Not refused.

## Stakes
High. Touches a SessionEnd hook that runs after every session on both hosts, and the shared craftsmanship-baseline ratchet for the whole repo.

## Baseline (.craftsmanship-baseline.json)
`git diff origin/main -- .craftsmanship-baseline.json` shows 2 entries removed, 0 added, only the generated date changed.

- `mcp_server/hooks/session_lifecycle.py`, method-size, `_run_consolidation`: real fix. origin/main's `_run_consolidation` was one function around 42 lines, doing turn-gating, mode selection, `asyncio.run`, and logging all inline. It is gone from the diff entirely, replaced by three functions each under 40 lines: `_consolidation_mode` (13 lines), `_spawn_consolidation` (around 37 lines, builds argv and calls Popen), `_run_consolidation_cycle` (around 30 lines, the body that used to run inline). Confirmed no function in the new file exceeds 40 lines.
- `mcp_server/hooks/session_lifecycle.py`, unsourced-constant, `_LONG_SESSION_TURNS`: real fix. origin/main has one `# source: ADR-0497` comment, oddly doubled, directly above `_SHORT_SESSION_TURNS` only; `_LONG_SESSION_TURNS` sits below with no source comment of its own. The diff gives each constant its own `# source: ADR-0497` line.

Zero additions confirmed. The baseline shrink is honest, not a regeneration that absorbs new debt.

## Rollout reader against real files
Ran `transcript_activity(path)` read-only against the three most recent `~/.codex/sessions/2026/09/*/rollout-*.jsonl` files.

File 1: tool_sequence length 246 (starting `exec`, `exec`, `exec`, `exec`, `spawn_agent`, ...), turn_count 39, matching an independently computed count of `response_item` records whose `payload.type == "message"` and `payload.role == "assistant"` (39) exactly. File 2: tool_sequence length 280 (`exec`, `wait`, ...), turn_count 58, independent count 58, match. File 3: tool_sequence length 0, turn_count 8, independent count 8, match.

Robustness: a copy of file 1 with its last line truncated mid-JSON plus a trailing blank line does not raise, and returns the same 246/39. An empty file and a missing path both return `{"tool_sequence": [], "turn_count": 0}` without raising.

Fixture versus real records, field by field: dumped the real rollout's first-seen record of every `(type, payload.type, payload.role)` shape and diffed against `codex_rollout_excerpt.jsonl`. `session_meta`, `event_msg`, `response_item/message/developer`, `response_item/message/user`, `response_item/message/assistant`, `world_state`, `turn_context`, `response_item/custom_tool_call` (name `exec`), `response_item/custom_tool_call_output`, `response_item/reasoning`, `response_item/function_call` (name e.g. `send_message` or `spawn_agent`), `response_item/function_call_output`, `token_usage_record`, `inter_agent_communication_metadata`, `response_item/agent_message`, `compacted`. Every shape the fixture models and every shape the docstring names as read-past is present in real data with the same field names.

## Claude path unchanged
`git diff origin/main -- mcp_server/infrastructure/transcript_activity.py` shows the original reader's body (open, iterate, substring-check `"assistant"`, `json.loads`, isinstance and type guards, tool-name extraction, cap, `except OSError`) is byte-identical, only relocated into `_claude_transcript_activity` under a new signature and docstring. `tests_py/infrastructure/test_transcript_activity.py` passes (21 of 21 combined with the new rollout tests). A whole-file-is-a-JSON-array transcript, where the parsed record is a `list` rather than a `dict`, still returns the empty result without raising: the `isinstance(record, dict)` guard flagged in the PR #592 review is present and unchanged.

## SessionEnd end to end
Isolated `HOME`, `CORTEX_CLAUDE_DIR`, and, after one wrong turn, `CORTEX_MEMORY_STORE_BACKEND=sqlite` plus `CORTEX_MEMORY_DB_PATH`. The first attempt set a nonexistent `CORTEX_DB_BACKEND=sqlite`; the real knob is `CORTEX_MEMORY_STORE_BACKEND` under the `CORTEX_MEMORY_` prefix, and the default is `STORE_BACKEND="auto"` with `DATABASE_URL="postgresql://127.0.0.1:5432/cortex"`. That first run's spawned `--consolidate light` child connected to a live local Postgres over TCP, confirmed via `sample`'s native call stack showing psycopg/libpq frames, and ran over 100 seconds before I killed it with `kill -9` once the cause was identified. Decay is the routine, idempotent-per-session operation this hook already runs on every real session end, so I judge the exposure low, but it is a testing mistake on my part worth flagging rather than omitting. The corrected run, with sqlite isolation, showed:

Piping a SessionEnd JSON event into `python -m mcp_server.hooks.session_lifecycle` exits 0 immediately; `session-log.json` gets the row with `sessionId`, `turnCount`, and `toolsUsed` filled from the transcript; stderr logs `consolidation spawned (mode=light) -> .../consolidate.log`. `ps -ef` during the run showed the spawned argv directly: `.../Python -m mcp_server.hooks.session_lifecycle --consolidate light`, with `ppid=1`, confirming `start_new_session=True` detachment, running separately from the parent hook process which had already exited. The child completes on its own: sqlite `memory.db` gets created and `consolidate.log` gets `Dream (light): 0 decayed, 0 compressed`, confirming `wire_composition_root()` ran in the child (the store import and DB creation only happen after wiring) before `_run_consolidation_cycle`.

Hang test: the parent's exit was already observed happening before the child, which ran over 100 seconds against Postgres in the mistaken first run, had done anything. That is direct proof the parent never waits on the child, independent of the unit test `test_the_session_log_row_exists_even_when_the_launcher_fails`, which pins the same order by making the launcher raise.

Non-blocking and unrelated to this diff: the child's `consolidate.log` logs four `wiki_*_pass failed (non-fatal): 'SqliteMemoryStore' object has no attribute 'batch_pool'` lines before the decay line. That is a pre-existing sqlite versus postgres parity gap in `consolidate_handler`'s wiki passes, not touched by this PR.

## Child interpreter
`python_executable()` returns `sys.executable`, the exact interpreter already running the hook. Since that interpreter is already executing `mcp_server.hooks.session_lifecycle` as `-m` at the moment it decides to spawn, the same `-m mcp_server.hooks.session_lifecycle --consolidate <mode>` is guaranteed importable in the same environment whether wheel-installed or not; no `scripts/` dependency exists.

## Concurrency
Ran two SessionEnds for different session_ids concurrently against the same isolated sqlite store, WAL mode, no explicit `timeout=` passed to `sqlite3.connect` so the sqlite3 module's default 5 second busy-timeout applies. Both completed. All three session-log rows from this test sequence were present, so the read-modify-write JSON log lost no update, and both consolidation cycles logged `Dream (light): 0 decayed, 0 compressed` with no "database is locked" error. Neither `consolidate_handler` nor the stores hold an explicit lock or advisory flag; concurrent writers rely on sqlite's own WAL single-writer serialization plus the connection's busy-timeout. On a real locked collision the failure mode is graceful: `_run_consolidation_cycle`'s `except Exception` catches `sqlite3.OperationalError` the same as any other handler failure and logs it non-fatal, with no corruption. This concurrency exposure predates this PR: two overlapping SessionEnd hook invocations already ran as two independent OS processes before this change, each calling `consolidate_handler` in-process. What changed is that the cycle now runs detached from its own hook process rather than blocking it, which if anything shortens the collision window per event rather than lengthening it. Not a finding against this PR.

## Mutation
Removing `start_new_session=True` from the `Popen` call fails `TestSpawnConsolidation::test_the_launcher_gets_the_same_mode_the_turn_count_earns` with `KeyError: 'start_new_session'`, 1 failed and 34 passed in `tests_py/hooks/test_session_lifecycle.py`.

Removing the `session_meta` dispatch, so `transcript_activity` always calls `_claude_transcript_activity`, fails `test_transcript_rollout.py::test_a_rollout_yields_its_tool_sequence_and_turn_count`: `tool_sequence` comes back `[]` instead of `['exec', 'exec', 'send_message', 'exec']`, 1 failed and 55 passed across the combined infra and hooks suites. Both files reverted cleanly afterward, `git diff --stat` empty.

## Static guards and gates
`pytest tests_py/scripts -q`: 1006 passed, 359 subtests passed. `python scripts/check_craftsmanship.py --base origin/main`: OK. `uvx ruff@0.16.6 check` and `ruff format --check` on all six changed Python files: clean. `pyright` on the three changed `mcp_server/` files: 0 errors, 0 warnings. Decision gate: ran `decision_gate._line_runs` directly against each changed file's real comment-line scan, which only counts `#`-prefixed lines via the language lexer and ignores docstrings; no file has a run of 8 or more consecutive comment lines, compliant with `PROSE_RUN_LIMIT`. ADR and wiki mirror parity: `docs/adr/ADR-1081-...md` and `docs/adr/ADR-1082-...md` diff byte-for-byte identical to `wiki/adr/cortex/1081-...md` and `1082-...md`, modulo the wiki generator's read-only-mirror banner line; `wiki/manifest.json` gained both `ADR-1081` and `ADR-1082` entries pointing at the right mirror filenames. Full `pytest -q`: 8893 passed, 8 skipped, 2 warnings, 459 subtests passed in 466.76s, matching the PR's claimed count exactly. `gh pr checks 610`: every non-skipped check passes, covering Analyze for actions/javascript-typescript/python, Build Package, CI Green, Changed paths, CodeQL, Craftsmanship Gate, Docker Smoke, Fuzz for address and undefined, HOL plugin scan, Lint, Release dependency set, Test across Python 3.10 through 3.13, Test on Windows with the SQLite backend, Test with the SQLite backend, Validate MCP host configurations, the upstream identity contract check, Type Check, and plugin-scanner. The three "skipping" jobs, Docker Build for devcontainer and runtime images and the scheduled Fuzz batch and Tag a release, are conditional jobs not triggered on this PR, not failures.

## Doc-claims finding
`CHANGELOG.md`'s Fixed entry for ADR-1082 says consolidation now runs in a detached `consolidate_launch` subprocess. No file, module, or symbol named `consolidate_launch` exists anywhere in the tree; grepping the whole repository for that string returns only that one CHANGELOG line. This is the PR's own report of a first design, a standalone `consolidate_launch.py` importing `mcp_server.handlers.consolidate`, that the craftsmanship gate refused as a new layer violation. The shipped design is the module re-invoking itself, `_spawn_consolidation` inside `session_lifecycle.py`, confirmed end to end above. The CHANGELOG line was left describing the rejected design rather than the one that shipped. ADR-1082 itself, its wiki mirror, and the code all agree with each other and correctly describe the self-relaunch; only the CHANGELOG prose drifted. Since the ADR, the authoritative decision record, is correct, and only the CHANGELOG needs a wording fix from "a detached `consolidate_launch` subprocess" to something like "a detached subprocess that re-invokes this module with `--consolidate <mode>`", this does not warrant REQUEST_CHANGES on its own. It still needs fixing before or immediately after merge: a changelog entry naming a module that was never written will mislead the next engineer who greps for it.

## Verdict
APPROVE, conditioned on fixing the `CHANGELOG.md` `consolidate_launch` line to describe the shipped self-relaunch design before or immediately after merge. Everything else measured clean: the baseline shrink is real and complete, the rollout reader is verified byte-for-byte against live Codex files with independently computed turn counts, the Claude path is provably untouched, the SessionEnd and consolidation split is verified end to end including the non-blocking spawn and detachment, both required mutants are killed by name, and every static gate plus the full suite and full CI are green.

